PR #4, opened 2019-05-20
Annetka k assignment1

annetka-k:
Done main assignment and 3 challenges


annetka-k/assignment-1/annetka-k-assignment-1/.gitattributes
```diff
@@ -0,0 +1,63 @@
+###############################################################################
```
ssancho2: These files don't need to be uploaded.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "anagram.h"
+#include <map>
+#include <vector>
+#include <string>
+#include <iostream>
+using namespace std;
+
+//deletes all the punctuation and whitespaces in string
+void strip(string& s);
+//breakes the string into strings
```
ssancho2: I think the comment is not very informative... breaks a sentence into words, maybe?

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+//deletes all the punctuation and whitespaces in string
+void strip(string& s);
+//breakes the string into strings 
+vector<string> split(const string& s);
+
+
+bool isAnagram(string s1, string s2) {
+	map<char, int> usedLetters;
+	for (int i = 0; i < s1.length(); i++) {
+		if (usedLetters.find(s1[i]) != usedLetters.end())
```
ssancho2: C++ has some 'magic' features that are useful here: If you access a map on a key that doesn't exist, it creates and adds the mapping with a default constructed value (0 for numbers), so you could replace lines 18-21 with just line 19!

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+//deletes all the punctuation and whitespaces in string
+void strip(string& s);
+//breakes the string into strings 
+vector<string> split(const string& s);
+
+
+bool isAnagram(string s1, string s2) {
+	map<char, int> usedLetters;
+	for (int i = 0; i < s1.length(); i++) {
```
ssancho2: Since you don't actually need the index, It's more idiomatic to use `for (char c : s1) { ... `

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+			return 0;
+		usedLetters[s2[i]]--;
+		if (usedLetters[s2[i]] == 0)
+			usedLetters.erase(s2[i]);
+	}
+	return usedLetters.empty();
+}
+
+bool isAnagramCaseIns(string s1, string s2) {
+	for (int i = 0; i < s1.length(); i++)
```
ssancho2: You can also use the iterator trick here, but using a reference instead of a copy: 
`for (char &c : s1) {
   c = tolower(c);
}`

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+		s1[i] = tolower(s1[i]);
+	for (int i = 0; i < s2.length(); i++)
+		s2[i] = tolower(s2[i]);
+	return isAnagram(s1, s2);
+}
+
+
+bool isAnagramSentence(string s1, string s2) {
+	strip(s1); strip(s2);
```
ssancho2: Most style guides are against having more that one statement in a line.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+	return isAnagramCaseIns(s1, s2);
+}
+
+bool isAnagramSentenceByWords(string s1, string s2) {
+	vector<string> v1(split(s1)), v2(split(s2));
+	if (v1.size() != v2.size())
+		return 0;
+	for (int i = 0; i < v1.size(); i++) {
+		if (!isAnagramCaseIns(v1[i], v2[i]))
+			return 0;
```
ssancho2: `return false` is a bit better.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return 1;
+}
+
+
+//Helper functions
+
+void strip(string& s) {
+	string news = "";
+	string delims = ",./?! -+)(*><;:";
```
ssancho2: c++ has a `ispunct` function.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/challenge1.cpp
```diff
@@ -0,0 +1,14 @@
+//Created by Anna Kravets
+#include <iostream>
+#include "anagram.h"
+using namespace std;
+
+int main1() {
```
ssancho2: Ideally, you want your test to be automated and not require visual inspection. The output should be the number of tests ran, and which ones failed. For the failed ones you want to see the expected answer and the actual answer.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+	for (int i = 0; i < s.length(); i++) {
+		if (delims.find(s[i]) == string::npos)
+			news += s[i];
+	}
+	s = news;
+}
+
+vector<string> split(const string & s) {
```
ssancho2: I think the idea in the problem was to ignore/remove punctuation first, then split into words.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <iostream>
+using namespace std;
+
+//deletes all the punctuation and whitespaces in string
+void strip(string& s);
+//breakes the string into strings 
+vector<string> split(const string& s);
+
+
+bool isAnagram(string s1, string s2) {
```
ssancho2: Parameter names can use some improvement :)

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -6,79 +6,91 @@
 #include <iostream>
 using namespace std;
 
-//deletes all the punctuation and whitespaces in string
-void strip(string& s);
-//breakes the string into strings 
-vector<string> split(const string& s);
+//breakes the sentence in words 
+vector<string> split(const string& sentence);
```
ssancho2: Some functions start with lower case and some with upper case. Pick one and be consistent :)

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -6,79 +6,91 @@
 #include <iostream>
 using namespace std;
 
-//deletes all the punctuation and whitespaces in string
-void strip(string& s);
-//breakes the string into strings 
-vector<string> split(const string& s);
+//breakes the sentence in words
```
ssancho2: NIT: It's a bit nicer to start the comment with a uppercase letter and leave a space:
// Breaks the sentence into words.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -6,79 +6,91 @@
...
 
-void strip(string& s) {
-	string news = "";
-	string delims = ",./?! -+)(*><;:";
-	for (int i = 0; i < s.length(); i++) {
-		if (delims.find(s[i]) == string::npos)
-			news += s[i];
-	}
-	s = news;
+string ToLower(const string& s) {
```
ssancho2: NIT: you can declare the parameter as `string s` and let the compiler do the copy for you :)

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -6,79 +6,91 @@
...
+string RemovePunctuation(const string& s) {
+  string newstr = "";
+  for (char c : s) {
+    if (!ispunct(c))
+      newstr += c;
+  }
+  return newstr;
+}
+
+string strip(const string& s) {
```
ssancho2: This function behaves very similar to RemovePunctuation, if there are internal functions you may have only one with an extra parameter like `bool removeSpaces` .
If you want to expose these functions, you can just call the private one like:
```
string RemovePuntuation(const string &sentence) {
   return RemovePunctuationInternal(sentence, false);
}
```

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/anagram.cpp
```diff
@@ -6,79 +6,91 @@
...
+  return newstr;
+}
+
+vector<string> split(const string & sentence) {
+  string sentenceWithoutPunctuation=RemovePunctuation(sentence);
+  vector<string> words;
+  int start = sentenceWithoutPunctuation.find_first_not_of(' '), end;
+  while (start != string::npos) {
+	  end = sentenceWithoutPunctuation.find_first_of(' ', start);
+	  if (end != string::npos)
```
ssancho2: I think the if is not needed (`npos` is the max unsigned int, so `npos - start` will be bigger that the string size, so `substr` will'll take all chars until the end).
Which code is clearer it's up for debate :), but probably better as it is.

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/main.cpp
```diff
@@ -1,14 +1,58 @@
...
-		cout << "not anagrams\n";
-	return 0;
-}
\ No newline at end of file
+#include "E:/C++/Group-10/shared/testing/assertions.h"
+#include "E:/C++/Group-10/shared/testing/test.h"
+
+namespace {
+
+  TEST(AnagramsWordsCaseSensitive) {
```
ssancho2: Great!!! :) 
A few more examples with empty strings and one character strings will be nice. Or strings in which the words have the same letters but different count, like "aabcd" and "abbcd".

annetka-k/assignment-1/annetka-k-assignment-1/annetka-k-assignment-1/main.cpp
```diff
@@ -1,14 +1,58 @@
...
+  }
+
+  TEST(AnagramsSentences) {
+    EXPECT_TRUE(areAnagramsSentences("debit card", "bad credit"));
+    EXPECT_TRUE(areAnagramsSentences("school master", "The classroom"));
+    EXPECT_TRUE(areAnagramsSentences("The eyes", "They see"));
+    EXPECT_TRUE(areAnagramsSentences("Funeral", "Real fun"));
+    EXPECT_TRUE(areAnagramsSentences("anagram", "nag a ram"));
+    //close but don't exactly match
+    EXPECT_FALSE(areAnagramsSentences("I am Lord Voldermort", "Tom Marvolo Riddle"));
```
ssancho2: :)

PR #5, opened 2019-05-20
Posiadacz07 - Assignment1

Posiadacz07:
Completed assignment 1:
* basic function to check if two strings are anagrams (basic task)
* checking both sensitive and insensitive case anagrams (challenge 1)
* function to check is sentences are anagrams ignoring white spaces and punctuation (challenge 2)
* all method tested and documented
* basic script to build and run tests


Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
+#include "Posiadacz07/assignment-1/anagrams.hpp"
+
+#include <algorithm>
+#include <map>
+
+// Helper function that checks if given char is lower case letter.
+bool isLowerCaseLetter(char character) {
```
ssancho2: c++ actually has a `islower` function. Your name is much better tough :)

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+// Helper function that checks if given char is upper case letter.
+bool isUpperCaseLetter(char character) {
+  return character >= 'A' && character <= 'Z';
+}
+
+// Helper function that checks if given char is number.
+bool isNumber(char character) { return character >= '0' && character <= '9'; }
+
+bool checkIfAnagrams(const std::string firstString,
```
ssancho2: `const std::string &` is a bit better, so you avoid making a copy that is not needed.

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+                     const std::string secondString) {
+  bool result = false;
+  std::map<char, int> alphabet;
+
+  for (const char &currentChar : firstString) {
+    if (isUpperCaseLetter(currentChar) || isLowerCaseLetter(currentChar) ||
```
ssancho2: There is also a `isalnum` function.

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+// Helper function that checks if given char is number.
+bool isNumber(char character) { return character >= '0' && character <= '9'; }
+
+bool checkIfAnagrams(const std::string firstString,
+                     const std::string secondString) {
+  bool result = false;
+  std::map<char, int> alphabet;
+
+  for (const char &currentChar : firstString) {
```
ssancho2: `const char &` is a bit overkill.

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+bool isUpperCaseLetter(char character) {
+  return character >= 'A' && character <= 'Z';
+}
+
+// Helper function that checks if given char is number.
+bool isNumber(char character) { return character >= '0' && character <= '9'; }
+
+bool checkIfAnagrams(const std::string firstString,
+                     const std::string secondString) {
+  bool result = false;
```
ssancho2: I think you don't need this variable, and removing it will improve readability.

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+      return result;
+    }
+  }
+  result = true;
+  return result;
+}
+
+bool checkIfAnagramsCaseInsensitive(const std::string firstString,
+                                    const std::string secondString) {
+  std::string firstStringToLowerCase = firstString;
```
ssancho2: It'll be easier to declare the parameters without the `const` and you get a copy for free.

Posiadacz07/assignment-1/anagrams_test.cpp
```diff
@@ -0,0 +1,104 @@
+#include "Posiadacz07/assignment-1/anagrams.hpp"
+
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+
+namespace {
+
+TEST(CheckTwoSameWords) {
```
ssancho2: Yay! Proper tests!!! :)

Posiadacz07/assignment-1/anagrams.cpp
```diff
@@ -0,0 +1,62 @@
...
+  std::string secondStringToLowerCase = secondString;
+
+  std::transform(firstStringToLowerCase.begin(), firstStringToLowerCase.end(),
+                 firstStringToLowerCase.begin(), ::tolower);
+  std::transform(secondStringToLowerCase.begin(), secondStringToLowerCase.end(),
+                 secondStringToLowerCase.begin(), ::tolower);
+  return checkIfAnagrams(firstStringToLowerCase, secondStringToLowerCase);
+}
+
+bool checkIfAnagramsCaseSensitive(const std::string firstString,
```
ssancho2: You could just rename `checkIfAnagrams` -> `checkIfAnagramsCaseSensitive` and remove this.

Posiadacz07/assignment-1/anagrams.cpp
Posiadacz07: Do you suggest to use it instead of those 3 functions `isUpperCaseLetter`, `isUpperCaseLetter` and `isNumber`?

Posiadacz07/assignment-1/anagrams.cpp
Posiadacz07: Will it be better to use just copy or const value?

Posiadacz07/assignment-1/anagrams.cpp
ssancho2: I think that it's better. Even if the name is horrible :) Maybe add a comment.
Using existing library functions where possible is a good idea. There is a balance when they are not exactly what you need, but here is kind of a perfect match.

Posiadacz07/assignment-1/anagrams.cpp
ssancho2: Just char. char is a basic type, making a copy is pretty cheap, and the compiler may be intelligent enough to put it into a register or optimize it somehow.

Posiadacz07/assignment-1/anagrams.cpp
ssancho2: The pros of reusing are that **usually** libraries are more robust, have strong tests, and have being tested in 'the real world' (sometimes for years).
Of course, if you try to reuse too much you may end up doing something like this: https://www.youtube.com/watch?v=X5SkW7K0e3Y :)

PR #6, opened 2019-05-27
Finish second assignment

sanjamijovic:
Completed the second assignment and provided tests.


sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -0,0 +1,213 @@
...
+}
+
+template <typename T>
+BinaryTree<T>& BinaryTree<T>::
+        insert(T value, T parent, bool left) noexcept(false) {
+    NodePtr parentNode = find(parent);
+
+    if ((left && parentNode->left != nullptr) ||
+        (!left && parentNode->right != nullptr)) {
+        throw InsertionException("Passed parent node already has two children");
```
ssancho2: The message it's not entirely correct, you may be trying to insert a second left child on a node that has no right child.
Also, Exceptions are generally not used on c++. They may be useful in a new project if everybody is on board with it, but it's hard to mix code that uses exceptions an exception-free code.

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -0,0 +1,213 @@
...
+    BinaryTree<T>& insert(T value, T parent, bool left) noexcept(false);
+
+    NodePtr find(T value) const;
+    std::list<NodePtr> findAncestors(T value) const;
+    NodePtr findCommonAncestor(T firstValue, T secondValue) const;
+
+private:
+    NodePtr root_;
+
+    void deleteSubtree(NodePtr root);
```
ssancho2: Using raw pointers it's hard :)
What happens with this code?
```
BinaryTree<int> tree1;
tree1 << 1 << 2 << 3;
BinaryTree<int> tree2 = tree1;
tree2 << 4 << 5;
```
What does tree1 have? What happens when the function ends?

sanjamijovic/assignment-2/test.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include "BinaryTree.h"
+
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+
+namespace {
+
+    TEST(GetNode) {
+        BinaryTree<int> tree = BinaryTree<int>();
+        tree << 7 << 3 << 4 << 2 << 5;
```
ssancho2: Love this!!! ^_^

sanjamijovic/assignment-2/test.cpp
```diff
@@ -0,0 +1,132 @@
+#include <iostream>
+#include "BinaryTree.h"
+
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+
+namespace {
+
+    TEST(GetNode) {
+        BinaryTree<int> tree = BinaryTree<int>();
```
ssancho2: I think this kind of the same as:
`BinaryTree<int> tree;`

sanjamijovic/assignment-2/test.cpp
```diff
@@ -0,0 +1,132 @@
+#include <iostream>
+#include "BinaryTree.h"
+
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+
+namespace {
+
+    TEST(GetNode) {
```
ssancho2: Yay, tests!!! :)
It may be good to add some tests for empty trees, tough.

sanjamijovic/assignment-2/test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        auto node = tree.find(100);
+        EXPECT_NULL(node);
+    }
+
+    TEST(FindAncestors) {
+        BinaryTree<int> tree = BinaryTree<int>();
+        tree << 7 << 3 << 4 << 2 << 5;
+        tree.insert(8, 4, false);
+        tree.insert(1, 2, true);
+        tree.insert(6, 2, false);
```
ssancho2: If you are going to reuse this tree it may be good to make a function to create it.

sanjamijovic/assignment-2/test.cpp
```diff
@@ -0,0 +1,132 @@
...
+        // root
+        ancestors = tree.findAncestors(7);
+        EXPECT_EQ(ancestors.size(), 0);
+
+        // element which is not contained in tree
+        ancestors = tree.findAncestors(100);
+        EXPECT_EQ(ancestors.size(), 0);
+    }
+
+    TEST(FindAncestorsCompleteTree) {
```
ssancho2: Nice!

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // level order insert
+    std::queue<NodePtr> queue;
+    queue.push(root_);
+
+    while (!queue.empty()) {
+        NodePtr current = queue.front();
+        queue.pop();
+
```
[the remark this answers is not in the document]
sanjamijovic: I read that the underlying array of the initializer list is an array of const objects. Since I need to change elements of the initializer list in the following block of code, I think this might not be a possible solution. I also tried to do it with `std::array` but the problem is that once I set the child's value it is not changing the pointer in the parent node when the child is null, even though the child is marked as a reference. Is there some possible elegant solution to this?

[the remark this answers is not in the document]
ssancho2: Yeah, you are right. That doesn't quite work.
You can do:
`for (Node<T> **childPtr : {&(current->left), &(current->right)}) { ... }`
But it's starting to get even more ugly :( Not sure it's worth it...

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -3,99 +3,85 @@
...
+        Node(T val) :
+                value(val),
+                left(nullptr),
+                right(nullptr),
+                parent(nullptr),
+                level(0)
+        {}
+
+        Node(T val, RawNodePtr par) :
```
ssancho2: NIT: par can use a better name, like parent.

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -3,99 +3,85 @@
...
-#include "InsertionException.h"
-
-template <typename T>
-struct Node;
 
 template <typename T>
 class BinaryTree {
-    typedef Node<T>* NodePtr;
+    struct Node;
+    typedef std::unique_ptr<Node> UniqueNodePtr;
```
ssancho2: So many types of pointers! :)

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -3,99 +3,85 @@
...
-
     if (root_ == nullptr) {
-        root_ = newNode;
+        root_.reset(new Node(value));
         return *this;
     }
 
-    // level order insert
-    std::queue<NodePtr> queue;
+    std::queue<std::reference_wrapper<UniqueNodePtr>> queue;
```
ssancho2: Since you use them only temporarily during this function, using raw pointers is fine here, and a bit simpler.

sanjamijovic/assignment-2/BinaryTree.h
```diff
@@ -111,68 +97,35 @@ BinaryTree<T> &BinaryTree<T>::operator<<(T value) {
...
-    if ((left && parentNode->left != nullptr) ||
-        (!left && parentNode->right != nullptr)) {
-        throw InsertionException("Passed parent node already has two children");
-    }
+BinaryTree<T>& BinaryTree<T>::insert(T value, T parent, bool left) {
+    auto parentNode = find(root_, parent);
 
-    auto newNode = new Node<T>(value, parentNode);
+    if (parentNode == nullptr)
+        return *this;
```
ssancho2: I think it's better to return `nullptr` on error, to make clear to the client that the insert didn't happen.
For that you need to change the return type from & to *, of course.

PR #7, opened 2019-05-28
added assignment 2 + testing

teodoraargintaru:


teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
+#ifndef HEADER_H
+#define HEADER_H
+
+#include <iostream>
+#include <bits/stdc++.h>
+#include <stdlib.h>
+
+using namespace std;
+
+typedef struct node{
```
ssancho2: Usually, struct/class names are defined with a name starting on a uppercase character, and you don't need the typedef, you can do:
```
struct Node {
...
};
```

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+    int val;
+    struct node *left, *right;
+}node;
+
+node *createNode(const int value){
+    node *newNode = new node;
+    newNode->val = value;
+    newNode->right = NULL;
```
ssancho2: `nullptr` is prefered to `NULL` (main reason being that `nullptr` has a pointer type, and `NULL` is just 0 :) )

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+    newNode->left = NULL;
+    return newNode;
+}
+
+void insert(node **root, const int value){
```
ssancho2: Where is the node inserted? Ideally, you should be able to create any tree, not only balanced ones.

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    queue<node*> q;
+    q.push(*root);
+
+    while(!q.empty()){
+        node *temp = q.front();
+        q.pop();
+
+        if (temp->left == NULL){
```
ssancho2: Not sure it's more clear, but instead of repeating code for left and right you can do:
`for (node *&child : {temp->left, temp->right}) { ... }`
or
`for (node **child : {temp->left, temp->right}) { ... }`

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+            q.push(temp->right);
+    }
+}
+
+// postorder traversal to find a node
+node *findNode(node *root, const int value){
+    if (root == NULL || root->val == value)
+        return root;
+
+    if (root->left) findNode(root->left, value);
```
ssancho2: You should keep the value returned and return it if it's not `nullptr`, right?

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+// the node with the searched key, so root is pushed in the ancestors
+// vector
+bool getAncestors(node *root, const int value, vector<int> &ancestors){
+    if (root == NULL) 
+        return false;
+    
+    if (root->val == value)
+        return true;
+
+    if (getAncestors(root->left, value, ancestors) || getAncestors(root->right, value, ancestors)){
```
ssancho2: NIT: keep it to 80 chars per line.

teodoraargintaru/assignment-2/header.h
```diff
@@ -0,0 +1,112 @@
...
+
+    return false;
+}
+
+// if both values are in either left or right subtree, then we have a common
+// ancestor; in this the function checks recursively if there is a lower
+// common ancestor than the value already found, or not
+// if neither of the right and left subtrees contain both values, then their
+// lowest common ancestor is root
+node *lowestCommonAncestor(node *root, const int value1, const int value2){
```
ssancho2: This looks like it's traversing the tree lot...
Also, what if one (or both) of the values are invalid?

teodoraargintaru/assignment-2/header.h
teodoraargintaru: Well if the node with the specific key is not found in the tree, I designed my function to return nullptr. I wrote like this instead of if the root is nullptr then return nullptr and another if for the other case.
Could you please explain why is this a problem exactly?

teodoraargintaru/assignment-2/header.h
teodoraargintaru: I was aware the line was too long, but I didn't know how to separate it so it would look nice afterwards.

Would something like this work?
if (getAncestors(root->left, value, ancestors) ||
    getAncestors(root->right, value, ancestors)){

If not, could you suggest a better option please?

teodoraargintaru/assignment-2/header.h
teodoraargintaru: Shouldn't the new node be inserted in the first empty space found?

ssancho2: That's the more common thing you may do, but then you can only create balanced trees. Ideally you want a generic data structure in which you can create any tree, for example, the tree in the problem statement is not balanced (the node 4 doesn't have a left child.)

teodoraargintaru/assignment-2/header.h
ssancho2: But if you call recursively to findNode, and it does find the node it will return it in here (line 55), but this function has to get that value and return it. You need something like:
```
if (root->left != nullptr) {
   node *ret = findNode(root->left, value);
   if (node != nullptr) return node;
}
if (root->right != nullptr) {
 ...
``

File: teodoraargintaru/assignment-2/header.h (hunk 6 above)
Comment (ssancho2) [reply teodoraargintaru/assignment-2/header.h]: Yes, that is fine.

File: teodoraargintaru/assignment-2/header.h [hunk 8]
```diff
@@ -97,16 +121,20 @@ bool isInSubtree(node *root, const int value){
 // if neither of the right and left subtrees contain both values, then their
 // lowest common ancestor is root
 node *lowestCommonAncestor(node *root, const int value1, const int value2){
-    if (root == NULL)
-        return NULL;
+    if (root == nullptr)
+        return nullptr;
     
-    if(isInSubtree(root->left, value1) && isInSubtree(root->left, value2))
+    if (findNode(root, value1) == nullptr ||
```
ssancho2: I like the function, it's very clear and concise, but you are still traversing the tree a lot :)
For example, this check can be done once at the beginning (in another function), no need to do on every recursive call...

teodoraargintaru/assignment-2/header.h
```diff
@@ -33,62 +33,86 @@ void insert(node **root, const int value){
...
     }
 
     return false;
 }
 
-// function that recursively checks if node with specific key can be found in
-// root's subtree
-bool isInSubtree(node *root, const int value){
-    if (root == NULL) 
+bool getAncestorsSafetyTest(node *root, const int value,
```
ssancho2: This function is not needed. If `getAncestors` doesn't find the node will return false on it's own.
Also, this is the main function your clients will use, if you actually need it, maybe it's better to call this one `getAncestors` and the other one `getAncestorsImpl` or similar.

teodoraargintaru/assignment-2/header.h
```diff
@@ -37,19 +37,35 @@ void insert(node **root, const int value){
...
+            } else{
+                q.push(*child);
+            }
+            
         }
     }
 }
 
+node *insertRight(node **root, const int value){
+    if (*root == nullptr){
```
ssancho2: On the `insertLeft` function, I mean that you don't needed this `*root == nullptr` special case, but you are right, the whole function was not really needed :)

PR #9, opened 2019-05-29
Posiadacz07 assignment2

Posiadacz07:
* basic implementation of binary tree with int key value
* find node function
* find ancestors function
* find lowest common ancestor function
* tests for implemented methods


Posiadacz07/assignment-2/BinaryTree_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+TEST(FindNodeInTree) {
+  BinaryTree tree = prepareBinaryTree();
+  BinaryTree::Node &findedNode = tree.findNode(5);
+  EXPECT_NOT_NULL(&findedNode);
+  EXPECT_EQ(5, findedNode.getKey());
+  EXPECT_NULL(findedNode.getLeftChild());
+  EXPECT_NULL(findedNode.getRightChild());
+}
+
+TEST(FindNodeNotExistingInTree) {
```
ssancho2: A few more tests will be nice... For example:
Using the functions on a empty tree.
Using non-existent keys.
Testing lowestCommonAncestor with one node being an ancestor of the other.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,106 @@
...
+  BinaryTree::Node *newNode = new BinaryTree::Node(key);
+  currentNode->setRightChild(newNode);
+  return newNode;
+}
+
+// Helper function to find node with given key or find ancestors of node with
+// given key.
+BinaryTree::Node *findNodeAndAncestors(BinaryTree::Node *currentRootNode,
+                                       std::vector<int> &ancestors, int key) {
+  if (!currentRootNode) {
```
ssancho2: It's a bit better to use `currentRootNode != nullptr`.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,106 @@
...
+    } else {
+      return nullptr;
+    }
+  }
+}
+
+BinaryTree::Node &BinaryTree::findNode(int key) {
+  BinaryTree::Node *rootNode = getRootNode();
+  std::vector<int> ancestors;
+  BinaryTree::Node *findNode = findNodeAndAncestors(rootNode, ancestors, key);
```
ssancho2: NIT: a bit wasteful of space...

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                               int key) {
+  BinaryTree::Node *newNode = new BinaryTree::Node(key);
+  currentNode->setRightChild(newNode);
+  return newNode;
+}
+
+// Helper function to find node with given key or find ancestors of node with
+// given key.
+BinaryTree::Node *findNodeAndAncestors(BinaryTree::Node *currentRootNode,
+                                       std::vector<int> &ancestors, int key) {
```
ssancho2: Usually, we try to avoid passing parameters by reference, for example we either use `const string&` or `string *`.
The main reason is that when you see a use of the function it's clear that it is/it may modify what you are passing., i.e:
```
findNodeAndAncestors(rootNode, ancestors, key);
```
vs
```
findNodeAndAncestors(rootNode, &ancestors, key);
```
It does has the drawback/feature that you may be getting a nullptr as parameter. In that case, you should check if ancestors is nullptr and not call `ancestors->push_back()` which can be used to save the wasted space on `findNode()`

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,106 @@
...
+}
+
+void BinaryTree::Node::setRightChild(Node *newRightChild) {
+  rightChild = newRightChild;
+}
+
+BinaryTree::Node *BinaryTree::getRootNode() const { return rootNode; }
+
+BinaryTree::Node *BinaryTree::setRootNode(int key) {
+  BinaryTree::Node *newNode = new BinaryTree::Node(key);
```
ssancho2: I see a lots of calls to `new` and no calls to `delete`... that's a bit scary.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,128 @@
...
+BinaryTree::Node *BinaryTree::insertRightChild(BinaryTree::Node *currentNode,
+                                               int key) {
+  BinaryTree::Node *newNode = new BinaryTree::Node(key);
+  currentNode->setRightChild(newNode);
+  return newNode;
+}
+
+// Helper function to find node with given key or find ancestors of node with
+// given key.
+BinaryTree::Node *findNodeAndAncestors(BinaryTree::Node *currentRootNode,
```
ssancho2: Good design making this function private!
Two minor comments:
* You could make this function a private method of Node, to avoid all `currentNode->` (but the parameter key and the member will clash, so need to rename one).
* It's ok here. But usually when a function comment says something like 'does this or/and that', or a class has multiple purposes, it's a hint that may be a need to split it.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,128 @@
...
+BinaryTree::Node *
+BinaryTree::lowestCommonAncestor(BinaryTree::Node *firstNode,
+                                 BinaryTree::Node *secondNode) {
+  if (firstNode == nullptr || secondNode == nullptr) {
+    return nullptr;
+  }
+  int firstKey = firstNode->getKey();
+  int secondKey = secondNode->getKey();
+
+  std::vector<int> firstAncestors = findAncestors(firstKey);
```
ssancho2: The problem statements says 'avoid copying/storing nodes in another container'. Those are not nodes, but it's close enough :)

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,128 @@
...
+  std::vector<int> secondAncestors = findAncestors(secondKey);
+
+  BinaryTree::Node *root = getRootNode();
+  int lowestCommonAncestorKey = root->getKey();
+
+  for (auto &firstAncestor : firstAncestors) {
+    for (auto &secondAncestor : secondAncestors) {
+      if (firstAncestor == secondAncestor &&
```
ssancho2: I see now that the problem statement is not very clear, but when we the problem statement says 'lowest', it means the node that is closest to the input nodes as possible (while still being an ancestor of both). (for example, the he 'highest common ancestor' is always root :) )

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,128 @@
...
+  if (currentNode == nullptr) {
+    return;
+  }
+  remove(currentNode->getLeftChild());
+  remove(currentNode->getRightChild());
+  delete currentNode;
+}
+
+void BinaryTree::Node::setLeftChild(Node *newLeftChild) {
+  if (leftChild != nullptr) {
```
ssancho2: NIT: you are already checking inside of `remove` is the pointer is `nullptr`. Since both functions are part of the same class, I don't think the double check is needed. It may be needed if you are calling a function in another library, for example.

Posiadacz07/assignment-2/BinaryTree.cpp
Posiadacz07: Ok, so what is your suggestion to split this function and avoid code duplication?

ssancho2: No, I think is fine in this case. If you split the function you will end up with two very similar ones.
Just a comment that in general comments of that type mean that.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -0,0 +1,118 @@
...
+  if (firstNode == nullptr || secondNode == nullptr) {
+    return nullptr;
+  }
+  int firstKey = firstNode->getKey();
+  int secondKey = secondNode->getKey();
+
+  BinaryTree::Node *root = getRootNode();
+  int lowestCommonAncestorKey = root->getKey();
+
+  for (auto &firstAncestor : findAncestors(firstKey)) {
```
ssancho2: Besides being O(n^2), this doesn't work when one of the nodes is an ancestor of the other (since the node is not returned as an ancestor of itself in that function, but should be considered one here).

Posiadacz07/assignment-2/BinaryTree_test.cpp
```diff
@@ -87,7 +87,7 @@ TEST(FindLowestCommonAncestorInTreeWithRoot) {
   BinaryTree::Node *secondNode = tree.findNode(6);
   BinaryTree::Node *lowestCommonAncestor =
       tree.lowestCommonAncestor(firstNode, secondNode);
-  EXPECT_EQ(7, lowestCommonAncestor->getKey());
+  EXPECT_EQ(3, lowestCommonAncestor->getKey());
```
ssancho2: 3 is not an ancestor of 7.

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -91,28 +91,51 @@ std::vector<int> BinaryTree::findAncestors(int key) {
...
+  }
+
+  if (rootNode->getKey() == firstNodeKey ||
+      rootNode->getKey() == secondNodeKey) {
+    return rootNode;
+  }
+
+  BinaryTree::Node *leftLowestCommonAncestor = nullptr;
+  BinaryTree::Node *rightLowestCommonAncestor = nullptr;
+  if (rootNode->getLeftChild() != nullptr) {
```
ssancho2: This is check is not needed since is the first thing you do on `lowestCommonAncestorImpl`, right?

Posiadacz07/assignment-2/BinaryTree.cpp
```diff
@@ -91,28 +91,51 @@ std::vector<int> BinaryTree::findAncestors(int key) {
   return ancestors;
 }
 
+// Helper function to find lowest common ancestor - traversing tree
+BinaryTree::Node *lowestCommonAncestorImpl(BinaryTree::Node *rootNode,
```
ssancho2: This function doesn't not handle invalid inputs, right? For example, if one on the keys is not in the tree it will return a pointer to the node with the valid key.

Posiadacz07/assignment-2/BinaryTree.cpp
Posiadacz07: Yes, you're right. It's not needed.

Posiadacz07/assignment-2/BinaryTree.cpp
Posiadacz07: Yes, so to be sure that input will be always right, will it be better to check if nodes exists in current tree? Try to find them? Probably in `lowestCommonAncestor` function to check it once.

ssancho2: As usual, there is a trade-of . Last year, I made the code that does the finding and checking at the same time (using a bitmask to return which nodes where found, or -1 in case of errors and a `T*` parameter to return the node value of the lca), and it was not pretty :) The code to go trough the tree twice to find each node, then once to find the common ancestor is the much easier to read and reason about.
In practice, I'll say write something that can be understandable even if it's not optimal. Optimize later, only if you have good reason to. Having two implementations can also be useful to test one against the other with random tests. (Google "premature optimization" :) )
On an interview, discuss it with the interviewer. Let them know you are considering both approaches and ask which one you should implement.

ssancho2: Ok, maybe I nerd sniped myself (https://xkcd.com/356/). I think I made the 'optimal' solution a bit prettier and easier to understand, but that bring us to another important point: "engineering time" :), that is another factor to have into account, and the reason we usually try to have "good enough" software. Also there is the saying that perfection is the enemy of progress, or something like that.

PR #11, opened 2019-06-11
Finish third assignment

sanjamijovic:
- Create AbstractLexicon class which provides lexicon interface and implements function for forming words present in lexicon from the given grid
  - Create SetLexicon class which implements a set representation of a lexicon
  - Create TrieLexicon class which implements a trie tree representation of a lexicon
  - Provide tests for all mentioned functionalities
  - Compare performances for forming words with different implementations of lexicon


sanjamijovic/assignment-3/AbstractLexicon.cpp
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    return formedWords;
+}
+
+void AbstractLexicon::formWordsRecursive(WordSet &words,
+        const AbstractLexicon::CharGrid &grid, int i, int j,
+        const std::string &currentWord,
+        AbstractLexicon::VisitedGrid visited) const {
```
ssancho2: Making a copy of the visited grid is expensive. Better to pass it as a pointer and ensure that when you return it has the same value that when you where called.  Something like:
```
{
   visited[i][j] = true;
   ...
   visited[i][j] = false;
}
```
This assumes that you check that `visited[i][j]` is `false` before calling the function.
And of course you also need to set to `false` before any return in the function.

sanjamijovic/assignment-3/AbstractLexicon.cpp
```diff
@@ -0,0 +1,58 @@
...
+        for (int j = 0; j < cols; j++) {
+            VisitedGrid visited(rows, std::vector<bool>(cols, false));
+            formWordsRecursive(formedWords, grid, i, j, "", visited);
+        }
+    }
+
+    return formedWords;
+}
+
+void AbstractLexicon::formWordsRecursive(WordSet &words,
```
ssancho2: It's a bit better to use a pointer instead of a reference, so it's clear when you are calling this function that the parameter will be modified.

sanjamijovic/assignment-3/AbstractLexicon.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+    int lowerRowBound = std::max(0, i - 1);
+    int upperRowBound = std::min(rows - 1, i + 1);
+    int lowerColumnBound = std::max(0, j - 1);
+    int upperColumnBound = std::min(cols - 1, j + 1);
+
+    for (int row = lowerRowBound; row <= upperRowBound; row++) {
+        for (int column = lowerColumnBound;
+             column <= upperColumnBound; column++) {
+            if ((row == i && column == j) || visited[row][column]) {
```
ssancho2: NIT: The check `(row == i && column == j)` is not needed, since `visited[i][j]` is `true`

sanjamijovic/assignment-3/AbstractLexicon.cpp
```diff
@@ -0,0 +1,58 @@
...
+            VisitedGrid visited(rows, std::vector<bool>(cols, false));
+            formWordsRecursive(formedWords, grid, i, j, "", visited);
+        }
+    }
+
+    return formedWords;
+}
+
+void AbstractLexicon::formWordsRecursive(WordSet &words,
+        const AbstractLexicon::CharGrid &grid, int i, int j,
```
ssancho2: `i` & `j` doesn't feel like clear names for the parameters. `y` & `x` or `row` & `col`/`column` sound better

sanjamijovic/assignment-3/SetLexicon.cpp
```diff
@@ -0,0 +1,26 @@
...
+    return *this;
+}
+
+bool SetLexicon::isWord(const std::string &word) const {
+    return words_.find(word) != words_.end();
+}
+
+bool SetLexicon::isPrefix(const std::string &prefix) const {
+    if (prefix.empty()) {
+        return false;
```
ssancho2: Why? I'll say that any word starts with an empty string :)
Does a test break if you remove this check? If not, you either need one more test or just remove it :)

sanjamijovic/assignment-3/TrieLexicon.cpp
```diff
@@ -0,0 +1,45 @@
...
+bool TrieLexicon::isWord(const std::string &word) const {
+    auto lastChar = getLastCharacter(word);
+    return lastChar != nullptr && lastChar->isEndOfWord;
+}
+
+bool TrieLexicon::isPrefix(const std::string &prefix) const {
+    return getLastCharacter(prefix) != nullptr;
+}
+
+TrieLexicon::TrieNode *TrieLexicon::getLastCharacter(
```
ssancho2: This doesn't return a character :)

sanjamijovic/assignment-3/TrieLexicon.h
```diff
@@ -0,0 +1,39 @@
...
+
+    bool isWord(const std::string& word) const override;
+    bool isPrefix(const std::string& prefix) const override;
+private:
+    TrieNodePtr root_;
+
+    TrieNode* getLastCharacter(const std::string& word) const;
+
+    struct TrieNode {
+        char character;
```
ssancho2: You don't use this, right?

sanjamijovic/assignment-3/performance.cpp
```diff
@@ -0,0 +1,170 @@
...
+            word += grid[currentRow][currentCol];
+
+            int lowerRowBound = std::max(0, currentRow - 1);
+            int upperRowBound = std::min(kNumRows - 1, currentRow + 1);
+            int lowerColumnBound = std::max(0, currentCol - 1);
+            int upperColumnBound = std::min(kNumCols - 1, currentCol + 1);
+
+            int newRow, newCol;
+
+            // it may get stuck in position where all neighbours are visited,
```
ssancho2: As an alternative, you could put all viable neighbors in a vector of pairs and pick one. Not sure which one is better.

sanjamijovic/assignment-3/performance.cpp
```diff
@@ -0,0 +1,170 @@
...
+            // so it will try to find unvisited neighbour at most 10 times
+            // (maximum number of neighbours in the grid + threshold because of
+            // the random function)
+            int numTries = 10;
+
+            do {
+                numTries--;
+                newRow = rand() / RAND_MAX *
```
ssancho2: Does this work? `rand() / RAND_MAX` is a integer division, and will almost always return 0...

sanjamijovic/assignment-3/SetLexicon.cpp
sanjamijovic: The test for empty lexicon will fall because of the implementation of the lower bound it will return that empty string is a prefix of the empty lexicon. Could you suggest what to do in that situation?

sanjamijovic/assignment-3/Grid.cpp
```diff
@@ -0,0 +1,83 @@
...
+    int rows = grid_.size();
+
+    for (int i = 0; i < rows; i++) {
+        int numCols = grid_[i].size();
+        visitedGrid->push_back(std::vector<bool>(numCols, false));
+    }
+}
+
+int Grid::numCols(int row) const {
+    if (grid_.size() < row || row < 0) {
```
ssancho2: `grid_.size() <= row `?

sanjamijovic/assignment-3/Grid.h
```diff
@@ -0,0 +1,38 @@
...
+public:
+    typedef std::vector<std::vector<char>> CharGrid;
+    typedef std::vector<std::vector<bool>> VisitedGrid;
+
+    inline char get(int row, int col) const {
+        return grid_[row][col];
+    }
+
+    inline bool isValid(int row, int col) const {
+        return row < grid_.size() && col < grid_[row].size();
```
ssancho2: Don't neet to check `row >=0` & `col >= 0`?

sanjamijovic/assignment-3/performance.cpp
```diff
@@ -147,22 +156,17 @@ void generateGridAndMatrixFiles() {
...
                 numTries--;
-                newRow = rand() / RAND_MAX *
-                         (upperRowBound - lowerRowBound) +
-                         lowerRowBound;
-                newCol = rand() / RAND_MAX *
-                         (upperColumnBound - lowerColumnBound) +
-                         lowerColumnBound;
-
-            } while (visited[newRow][newCol] && numTries > 0);
+            } while (visited[neighbour.first][neighbour.second] &&
```
ssancho2: You can check if that cell is visited visited before putting it in the array, so you don't need this loop .
You still need to check if there is any candidate.

sanjamijovic/assignment-3/test.cpp
```diff
@@ -28,14 +29,56 @@ namespace {
...
+
+        firstRow.push_back('A');
+        firstRow.push_back('R');
+
+        secondRow.push_back('T');
+        secondRow.push_back('C');
+        secondRow.push_back('P');
+        secondRow.push_back('D');
+
+        matrix.addRow(firstRow);
```
ssancho2: I think you can do:
`matrix.addRow({'A', 'R'});`

sanjamijovic/assignment-3/Grid.h
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    int numCols(int row) const;
+
+    Grid& addRow(std::vector<char>& row);
+
+    AbstractLexicon::WordSet formWords(const AbstractLexicon& lexicon);
+    void formVisited(VisitedGrid* visitedGrid);
```
ssancho2: This function could be private, right?

sanjamijovic/assignment-3/SetLexicon.cpp
ssancho2: Right, isPrefix should return false for the empty string if the lexicon is empty. 
I didn't see before, but bellow you need to check if `firstGreaterWord != words_.end()`. `lower_bound()` returns `end()` when the lexicon is empty or the prefix is greater than all words in the lexicon.

sanjamijovic:
> Very good seeing an implementation of a Trie! That is indeed the optimal (time-wise) data structure for the problem. How did they compare in the tests you ran? How many words there are in the lexicon on the random test?

I ran the performance test with 1000 words in the lexicon with the grid dimensions 100x80. The Set implementation was approx 1.3 times slower than Trie implementation (16609 ms for the Set implementation and 12816 ms for the Trie implementation).

PR #12, opened 2019-06-11
Assignment 3 - Posiadacz07

Posiadacz07:
* lexicon class
* wordSearch function
* basic tetsing


Posiadacz07/assignment-3/wordSearch.cpp
```diff
@@ -0,0 +1,69 @@
...
+  candidate.push_back(lettersGrid[i][j]);
+  isVisited[i][j] = true;
+
+  if (lexicon.isWord(candidate)) {
+    results.insert(candidate);
+  }
+
+  if (lexicon.isPrefix(candidate)) {
+    for (int direction = 0; direction < 8; ++direction) {
+      int nextI = i, nextJ = j;
```
ssancho2: Not sure if the function adds clarity, compare:
```
      int nextI = i, nextJ = j;
      nextGrid(nextI, nextJ, direction);
```
vs
```
      int nextI = i + row_offset[direction];
      int nextJ = j + column_offset[direction];
```

Posiadacz07/assignment-3/wordSearch.cpp
```diff
@@ -0,0 +1,69 @@
...
+  }
+  candidate.push_back(lettersGrid[i][j]);
+  isVisited[i][j] = true;
+
+  if (lexicon.isWord(candidate)) {
+    results.insert(candidate);
+  }
+
+  if (lexicon.isPrefix(candidate)) {
+    for (int direction = 0; direction < 8; ++direction) {
```
ssancho2: Better to use a constant for the number of directions, can be defined together with the arrays either as:
`const int num_directions = 8;`
 or
`const int num_directions = sizeof(row_offset) / sizeof(row_offset[0]);`

Posiadacz07/assignment-3/wordSearch.cpp
```diff
@@ -0,0 +1,69 @@
+#include "Posiadacz07/assignment-3/wordSearch.hpp"
+
+// Offset for calculating 8 directions: up, up_right, right, down_right, down,
```
ssancho2: For the case of moving on the eight directions, two nested loops from -1 to 1 may be easier. Not a big deal tough.

Posiadacz07/assignment-3/wordSearch.cpp
```diff
@@ -0,0 +1,69 @@
...
+  isVisited[i][j] = false;
+}
+
+std::set<std::string>
+wordSearch(const std::vector<std::vector<char>> &lettersGrid,
+           const Lexicon &lexicon) {
+  std::set<std::string> results;
+  int rows = lettersGrid.size();
+  if (rows > 0) {
+    int columns = lettersGrid[0].size();
```
ssancho2: Rows may have different sizes.

Posiadacz07/assignment-3/wordSearch.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Helper function that returns the next grid moving in the given direction
+void nextGrid(int &i, int &j, int direction) {
+  i += row_offset[direction];
+  j += column_offset[direction];
+}
+
+// Recursive function that traverse the grid of letters to form words
+void backTracking(const Lexicon &lexicon,
```
ssancho2: This function has quite a lot of parameters, that makes the function harder to read (and it's prone to errors).
One way to solve it, is to create a class to hold the state that doesn't change much during a exploration and make this function a member of it. `lexicon`, `lettersGrid`, `rows` & `columns` are great candidates for that, but you could also add `isVisited`, and even `results`. Instances of this class could be created to do one exploration (a call to `wordSearch`) and destroyed afterwards.
Also, `i` & `j` are names that may be confusing. Better names may be `row` & `col`/`column` or `y` & `x`

PR #13, opened 2019-06-12
solved assignment3

teodoraargintaru:


teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+#include <bits/stdc++.h>
+#include <stdlib.h> 
+
+#define SIZE 26
+
+using namespace std;
+
+// in solving the assignment, i assumed that all the letters in the
+// grid are uppercase letters
+typedef struct Grid{
```
ssancho2: This is not needed anymore, if you define:
```
struct Grid {
...
};
```
You already define the Grid class (the `struct` and `class` keywords work very similar, the only difference being that `struct` members are public by default and `class` members are private by default).

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
+#ifndef LEXICON_H
+#define LEXICON_H
+
+#include <iostream>
+#include <bits/stdc++.h>
+#include <stdlib.h> 
+
+#define SIZE 26
```
ssancho2: `SIZE` is not the best name for the number of letters in the alphabet :) 
Also, `#define` is usually discouraged for new code, better:
`const int kNumLetters = 26;`
or even
`const int kNumLetters = 'Z' - 'A' + 1;`

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+
+
+class Lexicon{
+public:
+    list<string> dictionary;
+    Grid myGrid;
+    vector<string> results;
+    Node *trie;
+    Lexicon();
+    Lexicon(vector<vector<char>> matrix, const int height, const int width,
```
ssancho2: `matrix` is being copied , better to declare it as `const vector<vector<char>> &matrix` if you don't need to change it, and `vector<vector<char>> *matrix` if you do.

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+        int &rightLimit, int &upperLimit, int &lowerLimit);
+
+private:
+};
+
+Lexicon::Lexicon(){
+    trie = new Node;
+}
+
+Lexicon::Lexicon(vector<vector<char>> matrix, const int height,
```
ssancho2: Can't you see `height` & `width` from `matrix`?

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+    const int width, list<string> dictionaryp){
+    dictionary = dictionaryp;
```
ssancho2: `dictionaryp` ?
Here we use the convention of using a trailing underscore for class members, that also helps with readability because you can see that something is a class member just from the name ending in '_'.

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+    myGrid.width = width;
+    myGrid.lettersMatrix = matrix;
+}
+
+bool Lexicon::isWord(const string word){
+    if (word.empty()){
+        return false;
+    }
+
+    if (find(results.begin(), results.end(), word) != results.end()){
```
ssancho2: Instead of lines 72 - 76, you could do:
`return find(results.begin(), results.end(), word) != results.end()`
find(results.begin(), results.end(), word) != results.end()

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+    const int width, list<string> dictionaryp){
+    dictionary = dictionaryp;
+    trie = new Node;
+    myGrid.height = height;
+    myGrid.width = width;
+    myGrid.lettersMatrix = matrix;
+}
+
+bool Lexicon::isWord(const string word){
+    if (word.empty()){
```
ssancho2: Is this check needed? The code bellow will give the correct answer anyway.
This will speed up things on the empty word query, but that it's done only once on the whole exploration, right?

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+
+    if (find(results.begin(), results.end(), word) != results.end()){
+        return true;
+    }
+    
+    return false;
+}
+
+bool Lexicon::isPrefix(const string word){
+    if (word.empty()){
```
ssancho2: I'll say that all words start with the empty string :)

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+        return false;
+    }
+
+    vector<string>::iterator it;
+    for (it = results.begin(); it != results.end(); ++it){
```
ssancho2: This is very expensive, if instead of a list of words you use a `set` (that has all elements in order, usually in a tree), you could do a binary search. Even better, you don't have to actually implement it, it's already there in `set::lower_bound` :)
You could do the same on a sorted `vector` with `std::lower_bound` from `<algorithm>`

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+        if (word.compare(it->substr(0, word.size())) == 0){
+            return true;
+        }
+    }
+
+    return false;
+}
+
+bool Lexicon::wordsToTrie(){
+    if (dictionary.size() == 0){
```
ssancho2: Not needed?

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    int index;
+    Node *auxiliaryTrie;
+        
+    for (string &word : dictionary){
+        transform(word.begin(), word.end(), word.begin(), ::toupper);
+        auxiliaryTrie = trie;
+
+        for (int i = 0; i < word.size(); i++){
```
ssancho2: You could do: `for (char c : word) { ... }`

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+            index = word[i] - 'A';
+            if(auxiliaryTrie->kids[index] == nullptr){
+                auxiliaryTrie->kids[index] = new Node;
+                auxiliaryTrie = auxiliaryTrie->kids[index];
```
ssancho2: NIT: common line `auxiliaryTrie = auxiliaryTrie->kids[index];` can be moved outside the if, and then you don't need a `else` block

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+        auxiliaryTrie->word = word;
+    }
+    return true;
+}
+
+void Lexicon::determineBoundaries(const int row, const int column,
+    int &leftLimit, int &rightLimit, int &upperLimit, int &lowerLimit){
+    (column > 0) ? leftLimit = column - 1 : leftLimit = column;
```
ssancho2: This is a bit confusing, maybe more clear as:
`leftLimit = column > 0 ?  column - 1 : column;`
or
`leftLimit = column > 0 ?  column - 1 : 0;`

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+            } else {
+                auxiliaryTrie = auxiliaryTrie->kids[index];
+            }
+        }
+        auxiliaryTrie->word = word;
+    }
+    return true;
+}
+
+void Lexicon::determineBoundaries(const int row, const int column,
```
ssancho2: `const int` it's a bit overkill...

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+        for (int j = leftLimit; j <= rightLimit; j++){
+            dfs(i, j, nextKid, resultsvec);
+        }
+    }
+    
+    myGrid.lettersMatrix[row][column] = letter;
+}
+
+void Lexicon::findWords(){
+    if (myGrid.height == 0 || myGrid.width == 0){
```
ssancho2: Not needed :)

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+        return;
+    }
+
+    vector<string> resultsvec;
```
ssancho2: You could clear `results` here and modify it inside `dfs`...

teodoraargintaru/assignment-3/assignment3/lexicon_test.cc
```diff
@@ -0,0 +1,83 @@
...
+
+    vector<vector<char>> matrix;
+    matrix.push_back({'R','A','T'});
+    matrix.push_back({'C','D','S'});
+    matrix.push_back({'A','O','G'});
+
+    Lexicon lexicon(matrix, matrix.size(), matrix[0].size(), dictionary);
+    lexicon.wordsToTrie();
+    lexicon.findWords();
+
```
ssancho2: Maybe add a test for isPrefix?

teodoraargintaru/assignment-3/assignment3/lexicon.h
```diff
@@ -0,0 +1,174 @@
...
+// in solving the assignment, i assumed that all the letters in the
+// grid are uppercase letters
+typedef struct Grid{
+        vector<vector<char>> lettersMatrix;
+        int height, width;
+    }Grid;
+
+typedef struct Node{
+    struct Node *kids[SIZE];
+    string word;
```
ssancho2: Usually, trie nodes only have a bool to indicate if it's a word. This is very expensive space-wise.

teodoraargintaru/assignment-3/assignment3/lexicon_test.cc
```diff
@@ -0,0 +1,83 @@
...
+    EXPECT_TRUE(lexicon.isWord("ARE"));
+    EXPECT_TRUE(lexicon.isWord("MERE"));
+    EXPECT_TRUE(lexicon.isWord("ANA"));
+    EXPECT_FALSE(lexicon.isWord("SI"));
+    EXPECT_FALSE(lexicon.isWord("PERE"));
+    EXPECT_FALSE(lexicon.isWord("MULTE"));
+}
```
ssancho2: No tests to see what you actually find in the grid? That it's in the results member, right?

teodoraargintaru/assignment-3/assignment3/lexicon.h
teodoraargintaru: If i declare it without typedef shouldn't I use struct Grid whenever I use this structure? Wouldn't it be more inconvenient?

teodoraargintaru/assignment-3/assignment3/lexicon_test.cc
teodoraargintaru: Sorry, I'm not sure I understand what you mean. If it's in the results member isn't it in the grid as well?

teodoraargintaru/assignment-3/assignment3/lexicon.h
ssancho2: I think all versions of c++ these days don't require that anymore, not sure when it changed.

teodoraargintaru/assignment-3/assignment3/lexicon_test.cc
ssancho2: Right, I haven't noticed that isWord & isPrefix are based on results.
Your implementation of the word finding is fine (more than fine, it's actually the optimal solution that I mentioned in the group meeting :) ), but the idea was to split it in two: One lexicon class to hold the list of words in the lexicon, implementing the functions isWord & isPrefix. One class/method that uses a Lexicon object and a grid to find words. Don't worry, It's fine.

PR #14, opened 2019-06-20
finished assignment 4

teodoraargintaru:


teodoraargintaru/assignment-4/assignment4/islands.h
```diff
@@ -0,0 +1,53 @@
...
+                        territory[row][col + 1] = 'S';
+                        toVisit.push({row, col + 1});
+                    } else if (col > 0 && territory[row][col - 1] == 'L'){
+                        territory[row][col - 1] = 'S';
+                        toVisit.push({row, col - 1});
+                    } else if (row > 0 && territory[row - 1][col] == 'L'){
+                        territory[row - 1][col] = 'S';
+                        toVisit.push({row - 1, col});
+                    } else {
+                        toVisit.pop();
```
ssancho2: Why do only one thing at a time? You can add all neighbors that are land and do `toVisit.pop()`

teodoraargintaru/assignment-4/assignment4/islands_test.cc
```diff
@@ -0,0 +1,49 @@
+#include "islands.h"
+#include "../../shared/testing/assertions.h"
+#include "../../shared/testing/test.h"
+
+TEST(ASSIGNMENT_EXAMPLE){
+    vector<vector<char>> territory;
+    territory.push_back({'S','L','S','L'});
```
ssancho2: No need to change now, but it could have been easier to create a function to convert `vector<string>` -> `vector<vector<char>>`, then you can do:
```
auto territory = convert(
  {"SLSL",
   "LLSS",
   "SSLS",
   "SSLS"};
```

teodoraargintaru/assignment-4/assignment4/islands_test.cc
```diff
@@ -0,0 +1,49 @@
...
+TEST(6x4){
+    vector<vector<char>> territory;
+    territory.push_back({'L','S','L','S'});
+    territory.push_back({'S','L','S','S'});
+    territory.push_back({'S','S','S','L'});
+    territory.push_back({'L','S','S','S'});
+    territory.push_back({'L','S','L','L'});
+    territory.push_back({'S','S','L','S'});
+    EXPECT_EQ(6, determineNumberOfIslands(territory));
+}
```
ssancho2: Other interesting cases to have are spirals, that will work fine with your code, but are 'harder' for bad algorithms.

teodoraargintaru/assignment-4/assignment4/n_dimension_territory.h
```diff
@@ -0,0 +1,63 @@
...
+            toVisit.push(i);
+            myMap.territoryType[i] = 'S';
+            islands++;
+
+            while(!toVisit.empty()){
+                int current = toVisit.top();
+                toVisit.pop();
+                
+                list<int>::iterator it;
+                for(it = myMap.neighbours[current].begin();
```
ssancho2: I think you can do: `for (int neighbour : myMap.neighbours[current]) { ... }`

teodoraargintaru/assignment-4/assignment4/n_dimension_territory.h
```diff
@@ -0,0 +1,63 @@
...
+using namespace std;
+
+class Map{
+    public:
+    Map();
+    Map(vector<list<int>> neighbours_, vector<char> territoryType_);
+    // function that does an iterative dfs in the graph
+    int numberOfIslands();
+    private:
+    // I designed my map as a graph; therefore it is assymetric
```
ssancho2: Nice!!!

teodoraargintaru/assignment-4/assignment4/n_dimension_territory.h
```diff
@@ -0,0 +1,63 @@
...
+
+Map::Map(){}
+
+Map::Map(vector<list<int>> neighbours_, vector<char> territoryType_){
+    myMap.neighbours = neighbours_;
+    myMap.territoryType = territoryType_;
+    myMap.numberOfTiles = territoryType_.size();
+}
+
+int Map::numberOfIslands(){
```
ssancho2: I think there is a little problem with the algorithm, what happens when:
```
territory = {'L','L','L'};
neighbors = {{2}, {2}, {}};
```
Maybe you need to enforce that neighbors is symmetric?

teodoraargintaru/assignment-4/assignment4/islands.h
```diff
@@ -0,0 +1,53 @@
+#ifndef ISLANDS_H
+#define ISLANDS_H
+
+#include<iostream>
+#include <bits/stdc++.h>
+
+using namespace std;
+
+int determineNumberOfIslands(vector<vector<char>> territory){
+    stack<pair<char,char>> toVisit;
```
ssancho2: Also, a BFS is usually better than a DFS for this, since the size of the `toVisit` structure is the perimeter of the already explored cells in a BFS, but can cover the whole area on a DFS. Think how the algorithms will work on a huge - all land board.

teodoraargintaru/assignment-4/assignment4/n_dimension_territory.h
```diff
@@ -47,11 +64,10 @@ int Map::numberOfIslands(){
                 toVisit.pop();
                 
                 list<int>::iterator it;
-                for(it = myMap.neighbours[current].begin();
-                    it != myMap.neighbours[current].end(); ++it){
-                    if(myMap.territoryType[*it] == 'L'){
-                        myMap.territoryType[*it] = 'S';
-                        toVisit.push(*it);
+                for(int neighbour : myMap.neighbours[current]){
```
ssancho2: `it` on the previous line is not needed.

teodoraargintaru/assignment-4/assignment4/n_dimension_territory.h
```diff
@@ -10,6 +10,8 @@ class Map{
     public:
     Map();
     Map(vector<list<int>> neighbours_, vector<char> territoryType_);
+    // function that enforces the neighbours vector to be symmetrical
+    void checkSymmetrical();
```
ssancho2: This should probably be private.

teodoraargintaru/assignment-4/assignment4/islands.h
```diff
@@ -7,16 +7,16 @@
...
     int height = territory.size();
 
     if(height == 0)
         return 0;
     
     int islands = 0;
-    int width = territory[0].size();
 
     for(int i = 0; i < height; i++){
+        int width = territory[i].size();
```
ssancho2: You also need to check bellow when you are moving up/down that the other row is as big as this one.
For example: `if (row < height - 1 && col < territory[row+1].size() && territory[row + 1][col] == 'L')`

PR #15, opened 2019-06-24
Solution for 2d maps with square tiles

annetka-k:


annetka-k/assignment-4/assignment4/islands.cpp
```diff
@@ -0,0 +1,39 @@
+#include "islands.h"
+
+const enum AreaType { SEA, LAND };
+
+
+bool outOfBounds(std::vector<std::vector<int>>* islandsMap, int row, int column) {
```
ssancho2: Nice, handling of jagged maps!

annetka-k/assignment-4/assignment4/main.cpp
```diff
@@ -0,0 +1,80 @@
+#include "islands.h"
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+
+// Pre: stringMap consists of '0' and '1'. '1' will be converted to 1, everything else to 0.
```
ssancho2: Why not use the `enum AreaType` ?

annetka-k/assignment-4/assignment4/main.cpp
```diff
@@ -0,0 +1,80 @@
...
+// Pre: numberOfRows divides the length of the string.
+// Creates a map of islands with particular number of rows.
+std::vector<std::vector<int>> createMap(std::string stringMap, int numberOfRows) {
```
ssancho2: Good idea to have an auxiliary function to do this. I think it will be a bit better to take a `vector<string>` instead of the two parameters, and you can replace 0 and 1 with something easier to see, like '.' & '#'. Then you can use it in all tests and is very readable, like:
```
  auto jagged = createMap({
    "#######",
    "....#",
    "##.##.#",
    "#.",
    "#.#",
    "#.#.#.#",
    "#.###"});
```

annetka-k/assignment-4/assignment4/main.cpp
```diff
@@ -0,0 +1,80 @@
...
+                                                {1, 1, 0, 0},
+                                                {0, 0, 1, 0},
+                                                {0, 0, 1, 0}};
+    std::vector<std::vector<std::vector<int>>> foundIslands;
+    EXPECT_EQ(3, countIslands(islandsMap, &foundIslands));
+    std::vector<std::vector<std::vector<int>>> foundIslandsExpected;
+    foundIslandsExpected.push_back(createMap("0100110000000000", 4)); 
+    foundIslandsExpected.push_back(createMap("0001000000000000", 4)); // second island
+    foundIslandsExpected.push_back(createMap("0000000000100010", 4)); // third island
+    EXPECT_EQ(3, foundIslands.size());
```
ssancho2: You can actually do `EXPECT_EQ(foundIslandsExpected, foundIslands);`

annetka-k/assignment-4/assignment4/main.cpp
```diff
@@ -0,0 +1,80 @@
...
+    for (int islandNum = 0; islandNum < foundIslands.size(); islandNum++) {
+      for (int row = 0; row < foundIslands[islandNum].size(); row++) {
+        for (int column = 0; column < foundIslands[islandNum][row].size(); column++)
+          EXPECT_EQ(foundIslandsExpected[islandNum][row][column], foundIslands[islandNum][row][column]);
```
ssancho2: Note that this doesn't help much with debugging, you will get an error like "expected 1, found 0", and you will be like "where!?!?" :)

annetka-k/assignment-4/assignment4/islands.cpp
```diff
@@ -0,0 +1,39 @@
...
+  return row < 0 || row >= islandsMap->size() || column < 0 || column >= islandsMap->at(row).size();
+}
+
+// If [row][column] is SEA or not valid then does nothing.
+// IF [row][column] is LAND then converts it to SEA and does the same for the 4 neighbours.
+// Post: adds LAND at [row][column] position in newIsland.
+void removeIsland(std::vector<std::vector<int>>* islandsMap, int row, int column, std::vector<std::vector<int>>* newIsland) {
```
ssancho2: A note on performance, a BFS is usually better than a DFS for this, since the size of the pending structure is the perimeter of the already explored cells in a BFS, but can cover the whole area on a DFS (in this case the pending structure is the stack). Think how the algorithms will work on a huge - all land board.
On the other hand, this is easier to implement, read and modify, so no need to hurry to implement something better unless needed.

PR #17, opened 2019-06-25
Finish fourth assignment

sanjamijovic:
- Create Rectangular map and implement method for finding number of
    islands
  - Create Arbitrary map where each tile can have arbitrary number of
    neighbours and implement method for finding number of islands in it
  - Implement method that returns all islands present in the Arbitrary
    map


sanjamijovic/assignment-4/ArbitraryMap.cpp
```diff
@@ -0,0 +1,91 @@
...
+    return newNode;
+}
+
+bool ArbitraryMap::addEdge(const NodePtr& first, const NodePtr& second) {
+    if (nodes_.find(first) == nodes_.end() ||
+        nodes_.find(second) == nodes_.end()) {
+        return false;
+    }
+
+    first->neighbours.insert(second);
```
ssancho2: Good you ensure that the graph is undirected, otherwise it's unclear what an island is (and harder to find)

sanjamijovic/assignment-4/ArbitraryMap.cpp
```diff
@@ -0,0 +1,91 @@
...
+    second->neighbours.insert(first);
+
+    return true;
+}
+
+int ArbitraryMap::numberOfIslands() const {
```
ssancho2: A note on performance, a BFS is usually better than a DFS for this, since the size of the pending structure is the perimeter of the already explored cells in a BFS, but can cover the whole area on a DFS (in this case the pending structure is the stack). Think how the algorithms will work on a huge - all land board.
On the other hand, this is easier to implement, read and modify, so no need to hurry to implement something better unless needed.

sanjamijovic/assignment-4/ArbitraryMap.cpp
```diff
@@ -0,0 +1,91 @@
...
+    for (const auto& node : nodes_) {
+        if (node->type == LAND) {
+            ArbitraryMap currentIsland;
+            std::unordered_map<NodePtr, NodePtr> oldToNewMapping;
+
+            if (visited.find(node) == visited.end()) {
+                dfsLandFind(node, &visited, &currentIsland, &oldToNewMapping);
+            }
+            if (currentIsland.numberOfTiles() != 0) {
+                islands.push_back(currentIsland);
```
ssancho2: NIT: you don't need the `if (currentIsland.numberOfTiles() != 0)` if you move this to the previous `if`, right?

sanjamijovic/assignment-4/ArbitraryMap.cpp
```diff
@@ -0,0 +1,91 @@
...
+        if (visited->find(neighbour) == visited->end()
+            && neighbour->type == LAND) {
+            dfsLandTraversal(neighbour, visited);
+        }
+    }
+
+}
+
+// returns maps representing islands present in current map
+std::vector<ArbitraryMap> ArbitraryMap::findIslands() {
```
ssancho2: Minor comment on the return type here. Since you return a series of `ArbitraryMap`, that are new, there is no way to relate them to the original map. For example, if there are several 1-cell islands, there will be several identical `ArbitraryMap` in the return, one for each.
To solve this, you could either add an id or some data to the cells, that you copy over to the new maps. Or instead of `ArbitraryMap` you can return a `vector` (or other structure) of the original nodes.

sanjamijovic/assignment-4/ArbitraryMap.h
```diff
@@ -0,0 +1,48 @@
...
+public:
+    enum TileType {
+        LAND,
+        SEA
+    };
+
+    NodePtr addNode(TileType type);
+    bool addEdge(const NodePtr& first, const NodePtr& second);
+
+    int numberOfIslands() const;
```
ssancho2: Note that the implementation of `numberOfIslands` & `findIslands` are very similar, maybe you don't need both? Or at least don't need both auxiliary functions `dfsLandTraversal` & `dfsLandFind`

sanjamijovic/assignment-4/RectangularMap.cpp
```diff
@@ -0,0 +1,46 @@
+#include "RectangularMap.h"
+
+int RectangularMap::numberOfIslands() const {
+    int numberOfIslands = 0;
+    std::vector<std::vector<bool>> visited(tiles_.size());
+
+    for (int row = 0; row < tiles_.size(); row++) {
+        visited[row] = std::vector<bool>(tiles_[row].size(), false);
```
ssancho2: Nice, handling jagged maps!

sanjamijovic/assignment-4/test.cpp
```diff
@@ -0,0 +1,127 @@
+#include "shared/testing/assertions.h"
+#include "shared/testing/test.h"
+#include "RectangularMap.h"
+#include "ArbitraryMap.h"
+
+namespace {
+    void createRectMap(std::vector<std::vector<enum RectangularMap::TileType>> *map) {
+        map->push_back({RectangularMap::TileType::SEA, RectangularMap::TileType::LAND,
```
ssancho2: No need to change now, but I think it will be easier to have a function that takes a vector<string> and returns `std::vector<std::vector<enum RectangularMap::TileType>>`. You can use 0 and 1 in the string or replace them with something easier to see, like '.' & '#'. Then you can use it in all tests and is very readable, like:
```
  auto jagged = createMap({
    "#######",
    "....#",
    "##.##.#",
    "#.",
    "#.#",
    "#.#.#.#",
    "#.###"});
```

PR #18, opened 2019-07-05
finished assignment5

teodoraargintaru:
My idea of solving this problem was approaching the dictionary as a graph, and then sort it in a topological order.


group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    adjacency[vertex1].push_back(vertex2);
+}
+
+bool Dictionary::isInconsistentDictionary(char vertex1, char vertex2){
+    list<char>::iterator it;
+    for(it = adjacency[vertex2].begin(); it != adjacency[vertex2].end(); ++it){
+        if (*it == vertex1){
+            return true;
+        }
+    }
```
sanjamijovic: It may be cleaner to use a foreach loop. Something like: `for (const auto& neighbor : adjacency[vertex2]) {...}`.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    map<char, int> indegree;
+};
+
+Dictionary::Dictionary(){
+}
+
+void Dictionary::addInAdjacencyList(char vertex1, char vertex2){
+    if(adjacency[vertex1].empty()){
+        list<char> foo;
+        adjacency[vertex1] = foo;
```
sanjamijovic: Why not `adjacency[vertex1] = list<char>();`?

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    // method that adds an empty adjacency list for the letters that
+    // have no constraints
+    void addRemainingLetters(vector<string> words);
+    void addInAdjacencyList(char vertex1, char vertex2);
+    // main method for the topological sorting algorithm, returns one 
+    // possible alphabet
+    vector<char> computeAlphabet();
+    // method that adds on the stack the letters on the 'inferior levels'
+    // after the ones on the 'superior levels'
+    void topologicalSortUtil(char letter, map<char, bool> &visited,
```
sanjamijovic: It may be better to make util functions private.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+#define DICTIONARY_H
+
+#include<iostream>
+#include<bits/stdc++.h>
+
+using namespace std;
+
+class Dictionary{
+    public:
+    bool isInconsistent = false;
```
sanjamijovic: It may be a bad thing returning a function status through a public field since it is exposed and can be changed. I would recommend passing a structure by pointer to function to store a result and returning bool as a status or making `isInconsistent` private and creating a getter for it.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    Dictionary();
+    Dictionary(vector<string> words);
```
sanjamijovic: Maybe it would be better to pass words by reference to avoid copying. Moreover, it can be a const reference.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+        
+        int j, k;
+
+        for(j = 0, k = 0; j < firstWord.size() && k < nextWord.size(); j++){
+            if(firstWord[j] != nextWord[k]){
+                if(isInconsistentDictionary(firstWord[j], nextWord[k])){
+                    isInconsistent = true;
+                }
+                addInAdjacencyList(firstWord[j], nextWord[k]);
+                indegree[nextWord[k]]++;
```
sanjamijovic: It is great that you compute input degree while building graph instead of doing it each time you call `computeAlphabet `.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+            k++;
+        }
+
+        addRemainingLetters(words);
+    }
+}
+
+Dictionary::Dictionary(vector<string> words){
+    N = words.size();
+    buildLists(words);
```
sanjamijovic: If `buildLists` is only used as util for this constructor I would consider making it private.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+
+    if(result.size() == adjacency.size()){
+        allPosibilities.push_back(result);
+    }
+}
+
+vector<vector<char>> Dictionary::computeAllAlphabets(){
+    vector<vector<char>> allPosibilities;
+
+    if(isInconsistent == true){
```
annetka-k: NIT: if(isInconsistent){...}

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
+#ifndef DICTIONARY_H
+#define DICTIONARY_H
+
+#include<iostream>
+#include<bits/stdc++.h>
+
+using namespace std;
```
Posiadacz07: There is better to not use ```using namespace ... ```. In big projects with a lot of namespaces there could be some mismatch.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+
+class Dictionary{
+    public:
+    bool isInconsistent = false;
+    Dictionary();
+    Dictionary(vector<string> words);
+    // method that builds the adjacency lists based on the first
+    // two letters that are different between two adjacent words in the
+    // dictionary
+    void buildLists(vector<string> words);
```
Posiadacz07: Better to use const reference or pointer to vector.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    // method that adds an empty adjacency list for the letters that
+    // have no constraints
+    void addRemainingLetters(vector<string> words);
```
Posiadacz07: Same as in line 17

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+
+Dictionary::Dictionary(){
+}
+
+void Dictionary::addInAdjacencyList(char vertex1, char vertex2){
+    if(adjacency[vertex1].empty()){
+        list<char> foo;
+        adjacency[vertex1] = foo;
+    }
+    adjacency[vertex1].push_back(vertex2);
```
annetka-k: I would suggest to check that we haven't already added vertex2 in adjacency[vertex1]. Just not to keep unnecessary parallel edges

group/assignment-5/assignment5/dictionary.h
Posiadacz07: It is not so nice to use pure reference to any of method arguments. When you want to modify them, use pointer.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+    vector<char> computeAlphabet();
+    // method that adds on the stack the letters on the 'inferior levels'
+    // after the ones on the 'superior levels'
+    void topologicalSortUtil(char letter, map<char, bool> &visited,
+        stack<char> &alphabet);
+
+    // OPTIONAL CHALLENGES-----------------------------------------------------
+    // method that backtracks all possible dictionaries by recursively removing
+    // all edges starting from vertices with indegree 0
+    void allTopologicalSortUtil(map<char, bool> &visited,
```
Posiadacz07: Same as in line 27

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+        vector<char> &result, vector<vector<char>> &allPossibilities);
+    // method that checks if vertex1 is already in the adjacency list
+    // of vertex2
+    bool isInconsistentDictionary(char vertex1, char vertex2);
+    // method that returns all possible dictionaries
+    vector<vector<char>> computeAllAlphabets();
+
+    private:
+    int N;
```
Posiadacz07: What this N value means? Maybe it will be better to rename it with something more descriptive?

group/assignment-5/assignment5/dictionary.h
Posiadacz07: What is more, foo is not very descriptive name.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+    return false;
+}
+
+void Dictionary::addRemainingLetters(vector<string> words){
+    for(int i = 0; i < N; i++){
+        for(int j = 0; j < words[i].size(); j++){
+            if(adjacency.find(words[i][j]) == adjacency.end()){
+                list<char> foo;
```
Posiadacz07: Same as in line 53

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+}
+
+void Dictionary::buildLists(vector<string> words){
+    for(int i = 0; i < N - 1; i++){
+        string firstWord = words[i];
+        string nextWord = words[i + 1];
+        
+        int j, k;
+
+        for(j = 0, k = 0; j < firstWord.size() && k < nextWord.size(); j++){
```
Posiadacz07: I don't see any usage of j and k values out of the loop, so you can remove lines 84 and 95 using:
```
for( int j = 0, k = 0; j < firstWord.size() && k < nextWord.size(); j++, k++) 
```
instead of this.

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+        alphabet.pop();
+    }
+    return result;
+}
+
+void Dictionary::allTopologicalSortUtil(map<char, bool> &visited,
+    vector<char> &result, vector<vector<char>> &allPosibilities){
+    map<char, list<char>>::iterator it;
+
+    for (it = adjacency.begin(); it!= adjacency.end(); ++it){
```
Posiadacz07: nitpick: to be consistent there should be spaces before and after !=

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+
+    if(isInconsistent == true){
+        return result;
+    }
+
+    map<char, bool> visited;
+    map<char, list<char>>::iterator it;
+    stack<char> alphabet;
+
+    for (it = adjacency.begin(); it!= adjacency.end(); ++it){
```
Posiadacz07: nitpick: same as in line 152

group/assignment-5/assignment5/dictionary.h
```diff
@@ -0,0 +1,197 @@
...
+        visited[it->first] = false;
+    }
+
+    for (it = adjacency.begin(); it!= adjacency.end(); ++it){
```
Posiadacz07: nitpick: same as in line 152

group/assignment-5/assignment5/dictionary_test.cc
```diff
@@ -0,0 +1,82 @@
...
+
+    return true;
+}
+
+TEST(ASSIGNMENT_EXAMPLE){
+    vector<string> words = {"ART", "RAT", "CAT", "CAR"};
+    Dictionary myDictionary(words);
+    vector<char> result = myDictionary.computeAlphabet();
+
+    bool case1 = vectorsAreEqual(result, {'T','A','R','C'});
```
Posiadacz07: It is a bit strange for me to use this kind of method instead of EXPECT_EQ. It works on vectors, so it will be more readable and you avoid using those case values and additional check with final value.

group/assignment-5/assignment5/dictionary_test.cc
```diff
@@ -0,0 +1,82 @@
...
+
+}
+
+TEST(INCONSISTENT_DICTIONARY){
+    vector<string> words = {"APPLE","PEAR","PEACH","ART"};
+    Dictionary myDictionary(words);
+    vector<char> results = myDictionary.computeAlphabet();
+
+    EXPECT_TRUE(myDictionary.isInconsistent);
+    EXPECT_TRUE(results.size() == 0);
```
Posiadacz07: You can use ```results.empty()``` or ```EXPECT_EQ(0, results.size())``` instead of operator==

group/assignment-5/assignment5/main.cc
```diff
@@ -0,0 +1,5 @@
+#include "dictionary.h"
```
Posiadacz07: This is probably not needed. Remove line 8 in build.sh and change line 6 to ```SRCS=($(echo *.cc= | sed -e 's/\<[a-z0-9_]*_test.cc\>//g'))```  and everything should works without this file.

PR #19, opened 2019-07-05
finished 5th assignment

annetka-k:
I consider a graph with letters as vertices and directed edges as following:
there is an edge from letter x to letter y if y>x according to the dictionary.
I do DFS on any unused letter and every time I finish DFS on a particular letter, I add it to a container.
Thus, the container has a property such that y is added earlier than x <=> x and y are unrelated or y can be reached from x in DFS bypass, which means y>x.


group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
...
+  char smallerLetter = pairOfLetters.first, biggerLetter = pairOfLetters.second;
+  if (smallerLetter!=biggerLetter)
+    letterNodes[smallerLetter]->biggerLetters.insert(biggerLetter);
+}
+
+
+Language::Node* Language::nextUnmarked(Node* current){
+  int currentIndex = 0;
+  if (current != nullptr)
+    currentIndex = current->letter+1;
```
sanjamijovic: Can be simplified to something like: `int currentIndex = (current != nullptr ? current->letter + 1 : 0);`

group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+
+std::pair<char, char> firstDifference(std::string smallerWord, std::string biggerWord) {
+  for (int i = 0; i < smallerWord.length() && i < biggerWord.length(); i++) {
+    if (smallerWord[i] != biggerWord[i])
+      return std::make_pair(smallerWord[i], biggerWord[i]);
+  }
+  return std::make_pair(biggerWord[smallerWord.length()], biggerWord[smallerWord.length()]);
+}
```
sanjamijovic: NIT: Just a style guide thing. Lines should be under 80 chars long to make the code more readable.

group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
+#include "alphabet.h"
+
+Language::Node::Node(char letter_) : letter(letter_), mark(UNSEEN) {}
+
+Language::Language() {
+  letterNodes.resize(256, nullptr);
+}
```
sanjamijovic: It may be space inefficient using a predefined size collection because we do not know the exact dimensions of the unknown alphabet. I understand that this is necessary for having predefined space for each char, but it may be better to use a map instead.

group/assignment-5/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+    "failure", "fear", "feel", "for", "from", "goal", "hand", "in", "inability", "independence", "instead", "interest", "irrational", "kilo", "lack", "lead", "learn", "little", "make",
+    "maladaptive", "may", "mentor", "methods", "mindset", "missteps", "mistakes", "neither", "nor", "not", "of", "on", "once", "openly",
+    "or", "other", "overcommitter", "path", "personal", "phenomenon", "poor", "psychological", "queen", "rather", "recurrent", "resilience", "responsibility",
+    "share", "speak", "struggle", "succeed", "success", "suffer", "syndrome", "tasks", "than", "that",  "the", "their", "them",  "these", "they", "this",
+    "tirelessly", "to", "type", "understand", "violon","when", "which", "who", "with", "wolf", "work", "worth", "xray","yarn", "zebra" };
+    Language language;
+    language.setUp(dictionary);
+    EXPECT_TRUE(language.findAlphabet(&alphabet));
+    EXPECT_EQ(alphabet, rightAlphabet);
+  }
```
sanjamijovic: Great test example!

group/assignment-5/main.cpp
```diff
@@ -0,0 +1,64 @@
...
+  TEST(AtrcOrTarc) {
+    std::vector<char> alphabet;
+    std::set<std::vector<char>> possibleAlphabets = { { 'a','t','r','c' }, { 't','a','r','c' } };
+    std::vector<std::string> dictionary = { "art","rat","cat","car" };
+    Language language;
+    language.setUp(dictionary);
+    EXPECT_TRUE(language.findAlphabet(&alphabet));
+    EXPECT_FALSE(possibleAlphabets.insert(alphabet).second);
```
sanjamijovic: It may be clearer to check something like: 
`EXPECT_NEQ(possibleAlphabets.count(alphabet), 0);`
Also in C++ 20 there is a `contains` function so we can do something like: `EXPECT_TRUE(possibleAlphabets.contains(alphabet));`

group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+
+// Post: first char in pair is smaller than second char.
+// Post: if there is not a valid pair returns pair of equal chars.
+std::pair<char, char> firstDifference(std::string smallerWord, std::string biggerWord);
+
+void Language::setUp(const std::vector<std::string>& dictionary) {
+  if (!dictionary.empty())
+    this->addLettersFromWord(dictionary[0]);
+  for (int i = 0; i < (int)dictionary.size() - 1; i++) {
```
teodoraargintaru: I don't think you really need to cast size_t to int here.

group/assignment-5/main.cpp
teodoraargintaru: I agree with @sanjamijovic. This example is impressive! :)

group/assignment-5/alphabet.cpp
Posiadacz07: For me it is ok. It is well known solution for problems when number of elements is limited like with characters. 
The elements of std::vector are stored contiguously that reduce random access complexity. Operations in map have logarithmic complexity.

group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+void Language::addLettersFromWord(std::string word) {
+  for (char letter : word)
+    if (letterNodes[letter] == nullptr)
+      letterNodes[letter] = new Node(letter);
+}
+
+void Language::insertPair(std::pair<char, char> pairOfLetters) {
+  char smallerLetter = pairOfLetters.first, biggerLetter = pairOfLetters.second;
+  if (smallerLetter!=biggerLetter)
```
Posiadacz07: nitpick: to be consistent there should be spaces before and after ```!=```

group/assignment-5/alphabet.cpp
```diff
@@ -0,0 +1,88 @@
...
+    letterNodes[smallerLetter]->biggerLetters.insert(biggerLetter);
+}
+
+
+Language::Node* Language::nextUnmarked(Node* current){
+  int currentIndex = 0;
+  if (current != nullptr)
+    currentIndex = current->letter+1;
+  while (currentIndex < 256) {
+    if (letterNodes[currentIndex] && letterNodes[currentIndex]->mark==UNSEEN)
```
Posiadacz07: nitpick: same as in line 64

group/assignment-5/alphabet.cpp
sanjamijovic: Right, agree with that.

PR #20, opened 2019-07-06
Finish fifth assignment.

sanjamijovic:
- implement alphabet as directed graph where nodes are letters and
    edges connect nodes based on their order
  - create function that stores all found alphabets from passed
    dictionary in set and returns an indicator whether passed
    dictionary corresponds to any consistent alphabet


group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+    nodes_[letter] = node;
+}
+
+bool AlphabetGraph::addEdge(char fromLetter, char toLetter) {
+    if (!containsLetter(fromLetter) || !containsLetter(toLetter)) {
+        return false;
+    }
+
+    auto fromNode = nodes_[fromLetter], toNode = nodes_[toLetter];
+    fromNode->outAdjacencyList.push_back(toNode);
```
annetka-k: Maybe we should check that toNode isn't already in the vector?  (In order to avoid parallel edges)

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+    return maxLength;
+}
+
+// Implements Khan's algorithm for topological sort adapted to return all
+// possible sorting alternatives. Returns an indicator whether passed dictionary
+// corresponds to any consistent alphabet and, if so, stores all corresponding
+// alphabets in passed set.
+bool AlphabetGraph::findAlphabets(std::set<std::vector<char>>
+                                  *allAlphabets) const {
```
annetka-k: I would suggest clearing allAlphabets at the beginning

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+            nodesToVisit.push(node);
+        }
+    }
+
+    // Goes over all possible current alternatives for forming alphabet. Repeats
+    // for the initial size of the queue, because nodes are being returned to
+    // the queue in order to be passed to recursive method in the next
+    // iteration. Otherwise, if this loop would iterate over all elements in
+    // queue it would be an infinite loop.
+    for (int i = 0; i < nodesToVisit.size(); i++) {
```
annetka-k: I think it would me more clear to say 
int numOfIterations=nodesToVisit.size(); 
and then use it in the loop.

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+                                currentAlphabet);
+
+        // returning nodesToVisit in initial state before the call, so that it
+        // will be valid for other calls for different alphabet alternatives, in
+        // order to use same collection instead of making copies
+        if (std::find(node->outAdjacencyList.begin(),
+                      node->outAdjacencyList.end(),
+                      nextToVisit) == node->outAdjacencyList.end()) {
+            nodesToVisit->push(nextToVisit);
+        }
```
annetka-k: I believe it would be more clear if we do 
nodesToVisit->push(nextToVisit);  without any check. 
And then later in the code we can make sure that we pop all nodes we have added before. By
for (const auto& neighbour : node->outAdjacencyList) {
    if (neighbour->inDegree++ == 0)
        nodesToVisit->pop();
  }

annetka-k: Maybe check whether it works for {"a","ba","bc"}?

group/assignment-5/AlphabetGraph.cpp
Posiadacz07: For me current naming is clear.

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+void AlphabetGraph::findAlphabetRecursively(NodePtr node,
+                                            std::queue<NodePtr> *nodesToVisit,
+                                            std::set<std::vector<char>> *allAlphabets,
+                                            std::vector<char> *currentAlphabet) const {
+    currentAlphabet->push_back(node->value);
+
+    if (currentAlphabet->size() == nodes_.size()) {
+        allAlphabets->insert(*currentAlphabet);
+
+        // returning currentAlphabet in initial state before the call, so that
```
Posiadacz07: nitpick: returns

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    for (int i = 0; i < nodesToVisit->size(); i++) {
+        NodePtr nextToVisit = nodesToVisit->front();
+        nodesToVisit->pop();
+
+        findAlphabetRecursively(nextToVisit, nodesToVisit, allAlphabets,
+                                currentAlphabet);
+
+        // returning nodesToVisit in initial state before the call, so that it
```
Posiadacz07: nitpick: same like 106 - returns

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+        // will be valid for other calls for different alphabet alternatives, in
+        // order to use same collection instead of making copies
+        if (std::find(node->outAdjacencyList.begin(),
+                      node->outAdjacencyList.end(),
+                      nextToVisit) == node->outAdjacencyList.end()) {
+            nodesToVisit->push(nextToVisit);
+        }
+    }
+
+    // returning inDegrees in initial state before the call, so that they will
```
Posiadacz07: nitpick: same as in 106 - returns

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+    // be valid for other calls for different alphabet alternatives
+    for (const auto& neighbour : node->outAdjacencyList) {
+        ++neighbour->inDegree;
+    }
+
+    // returning currentAlphabet in initial state before the call, so that it
```
Posiadacz07: nitpick: same as in 106 - returns

group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,146 @@
...
+    NodePtr node = std::make_shared<Node>(Node(letter));
+    nodes_[letter] = node;
+}
+
+bool AlphabetGraph::addEdge(char fromLetter, char toLetter) {
+    if (!containsLetter(fromLetter) || !containsLetter(toLetter)) {
+        return false;
+    }
+
+    auto fromNode = nodes_[fromLetter], toNode = nodes_[toLetter];
```
teodoraargintaru: Your naming style is very clear! :)

group/assignment-5/test.cpp
```diff
@@ -0,0 +1,59 @@
...
+                                               "PAN", "PLANT",
+                                               "TAN", "TAP",
+                                               "ANT",
+                                               "LAP"};
+        std::set<std::vector<char>> expected = {{'N', 'P', 'T', 'A', 'L'}};
+        AlphabetGraph graph(dictionary);
+
+        std::set<std::vector<char>> alphabet;
+        EXPECT_TRUE(graph.findAlphabets(&alphabet));
+        EXPECT_TRUE(alphabet == expected);
```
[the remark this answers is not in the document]
sanjamijovic: The problem is that `EXPECT_EQ` internally uses the operator `!=` which is not overloaded for the `std::set`.

PR #21, opened 2019-07-07
Posiadacz07 - finished assignment 5

Posiadacz07:
* basic function to find alphabet
* basic testing
* basic documentation


group/assignment-5/alienOrder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    int wordSecondLength = wordSecond.length();
+    int length = std::max(wordFirstLength, wordSecondLength);
+
+    for (int j = 0; j < length; j++) {
+      if (j < wordFirstLength && graph.find(wordFirst[j]) == graph.end()) {
+        graph[wordFirst[j]] = std::unordered_set<char>();
+      }
+      if (j < wordSecondLength && graph.find(wordSecond[j]) == graph.end()) {
+        graph[wordSecond[j]] = std::unordered_set<char>();
+      }
```
annetka-k: Maybe it would look simpler if we create a separate function to add letters from a word in the graph?

group/assignment-5/alienOrder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  }
+  return graph;
+}
+
+//! Helper function to compute degrees for letters
+std::unordered_map<char, int> computeInDegree(const GraphType &graph) {
+  std::unordered_map<char, int> degrees;
+  for (const auto &adj : graph) {
+    if (!degrees[adj.first]) {
+    }
```
annetka-k: What is this check for?

group/assignment-5/alienOrder.hpp
```diff
@@ -0,0 +1,13 @@
...
+#define GROUP_ASSIGNMENT5_ALIEN_ORDER
+
+#include <string>
+#include <vector>
+
+//! Basic function that search for alphabet for given words.
+//! param[in] words collection of words to create alphabet.
+//! returns   string that conteins right order of letters in alphabet.
+//!           If dictionary is inconsistent returns empty string.
+std::string findAlienOrder(std::vector<std::string> &words);
```
annetka-k: Why not const &?

group/assignment-5/alienOrder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  for (int i = 0; i < numNodes; i++) {
+    if (toVisit.empty()) {
+      return "";
+    }
+    char c = toVisit.front();
+    toVisit.pop();
+    order += c;
+
+    for (char neigh : graph[c]) {
+      if (!--degrees[neigh]) {
```
sanjamijovic: NIT: It may be clearer to check `if (--degrees[neigh] == 0) {...}`

group/assignment-5/alienOrder.cpp
sanjamijovic: I agree with @annetka-k. I think it would be much cleaner. Another way we can make the code more intuitive is to iterate over all letters in the first word outside this loop and create sets for them in the graph, then to the same with the second word. However, it would be a bit less efficient.

group/assignment-5/alienOrder.cpp
```diff
@@ -0,0 +1,83 @@
...
+      degrees[neigh]++;
+    }
+  }
+  return degrees;
+}
+
+std::string findAlienOrder(std::vector<std::string> &words) {
+  if (words.size() == 1) {
+    return words[0];
+  }
```
sanjamijovic: What about words with the repeating characters? E.g. `banana` should return `ban` or something like that.

group/assignment-5/alienOrder.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  GraphType graph = makeGraph(words);
+  std::unordered_map<char, int> degrees = computeInDegree(graph);
```
sanjamijovic: I think it may be better to implement the graph representation in an object-oriented way. That way we can separate the construction of the graph and `findAlienOrder`. The graph can be reused for more function calls. The information about the input degree could be stored in the node and then we do not need to compute input degrees for each call of `findAlienOrder`. If you still prefer a procedural solution than you can replace raw `char` type in `GraphType` with the node structure which also stores information about input degree. That way degree of each node can be computed it in `makeGraph` function to make it a bit more efficient.

group/assignment-5/alienOrder_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace {
+TEST(EmptyDictionary) {
+  std::vector<std::string> dictionary;
+  EXPECT_EQ(0, findAlienOrder(dictionary).length());
+}
+
+TEST(ConsistentDictionaray) {
+  std::string alphabet = "atrc";
+  std::vector<std::string> dictionary = {"art", "rat", "cat", "car"};
+  EXPECT_EQ(alphabet, findAlienOrder(dictionary));
```
sanjamijovic: Since there are two possible solutions (`atrc` and `tarc`), it may be better to check if the returned order corresponds to any of them. If something is changed in the implementation, the test may fail because the test expects one solution and the function may return the other.

group/assignment-5/alienOrder_test.cpp
```diff
@@ -0,0 +1,33 @@
...
+  std::vector<std::string> dictionary = {
+      "a",
+      "x",
+      "b",
+      "c",
+  };
+  EXPECT_EQ(findAlienOrder(dictionary), alphabet);
+}
+
+TEST(InconsistentDictionary) {
```
teodoraargintaru: Maybe add a test for an empty initial string, as well?

PR #22, opened 2019-07-09
Group assignment5

sanjamijovic:
Correct the code based on reviews


group/assignment-5/AlphabetGraph.cpp
```diff
@@ -0,0 +1,155 @@
...
+        // order to use same collection instead of making copies
+        nodesToVisit->push(nextToVisit);
+    }
+
+    // returns inDegrees and nodesToVisit in initial state before the call,
+    // so that they will be valid for other calls for different alphabet
+    // alternatives
+    for (const auto& neighbour : node->outAdjacencyList) {
+        if (neighbour->inDegree++ == 0) {
+            nodesToVisit->pop();
```
annetka-k: I believe there we actually intended to delete elements from the back of the queue. 
(Because earlier we pushed neighbours to the back and we assume that the previous loop ( for (int i=0; ... ) hasn't changed the nodesToVisit) 
So, it seems to me that it would be better to use deque instead of queue and switch to pop_back() method there

sanjamijovic: Right, switched to deque

PR #23, opened 2019-07-17
finished main task of 6th assignment

teodoraargintaru:
I uploaded my attempt for the main task of the 6th assignment (no extra challenges tackled...yet).
My idea: represent the parking lot as an array, where numbers from 1...N-1 (N is the size of the array) are cars, and 0 is the empty spot. Then all the cars that are not where they should be are pushed in a queue. 
As long as the queue is not empty, the car from the top of the queue is moved to the empty spot, and pushed again in the queue if it's still not where it should be.


group/assignment-6/assignment6/cars_test.cc
```diff
@@ -0,0 +1,40 @@
...
+        16,14,23, 1,33,88,39,70,100,47,82,62,60,38,13,7,59,52,18,43,67,81,79,
+        29,22,80,71,83,94,36,86,97,45,42,74,3,31,61,73,57,91,87,30,75,15,40,84,
+        19,9,65,89,46,32,44,2,51,68,63,69};
+    vector<int> final = {99,33,66,70,47,95,69,86,78,91,97,45,35,84,48,40,42,49,
+        82,94,50,31,88,11,28,68,89,46,81,16,4,63,76,8,24,34,2,29,21,3,55,58,19,
+        10,75,74,79,13,65,5,92,72,18,77,12,100,38,53,27,93,37,36,43,85,7,61,64,
+        41,57,17,30,80,87,56,25,39,67,9,51,98,14,6,44,60,54,59,73,1,83,52,62,32,
+        96,20,26,23,22,71,0,90,15};
+    vector<vector<int>> result = generateParkingSeq(initial, final);
+    EXPECT_EQ(result[result.size() - 1], final);
```
sanjamijovic: Great that you provided a bigger test!

group/assignment-6/assignment6/cars.h
```diff
@@ -0,0 +1,65 @@
+#ifndef CARS_H
+#define CARS_H
+
+#include<iostream>
+#include<bits/stdc++.h>
+
+using namespace std; 
+
+// returns position of a car in the parking
+int getPosition(const vector<int> parkedCars, const int car){
```
annetka-k: Maybe const vector<int>& ?

group/assignment-6/assignment6/cars.h
```diff
@@ -0,0 +1,65 @@
...
+    for(int i = 0; i < parkedCars.size(); i++){
+        if(parkedCars[i] == car){
+            return i;
```
annetka-k: Probably, it would be better to have pos instead of i.

group/assignment-6/assignment6/cars.h
```diff
@@ -0,0 +1,65 @@
...
+vector<vector<int>> generateParkingSeq(const vector<int> initialOrder,
+    const vector<int> finalOrder){
+    vector<vector<int>> result;
+    result.push_back(initialOrder);
+    queue<int> toBeArranged;
+
+    insertCarsNotOk(toBeArranged, initialOrder, finalOrder);
+
+    while(!toBeArranged.empty()){
+        vector<int> current = result[result.size() - 1];
```
annetka-k: We could also use result.back() there.

group/assignment-6/assignment6/cars.h
```diff
@@ -0,0 +1,65 @@
...
+    parkedCars[fromPos] = 0;
+}
+
+// inserts in queue all the cars that are not where their final position
+// should be
+void insertCarsNotOk(queue<int> &toBeArranged, const vector<int> initialOrder,
+    const vector<int> finalOrder){
+    for(int i = 0; i < initialOrder.size(); i++){
+        int car = initialOrder[i];
+        if(car != 0 && getPosition(initialOrder, car) !=
```
annetka-k: getPosition(initialOrder, car) will be anyway equal to i, right?
Also, i -> pos?

group/assignment-6/assignment6/cars.h
teodoraargintaru: I completely agree. It would be more elegant to use result.back().

PR #24, opened 2019-07-19
assignment6 - main

annetka-k:
I represent parking as a map with lots' ID as keys. Car ID and lot ID could be any integer. To rearrange cars I take two steps:
1. If currently empty lot is intended to have a car, then I will move the intended car to the empty lot.
2. Else I will choose any car that is on the wrong parking lot and move it to the currently empty lot.
The obtained sequence of moves will be of the minimum possible size.


group/assignment-6/assignment-6/cars.cpp
```diff
@@ -0,0 +1,99 @@
...
+  for (auto pair : plannedCars) {
+    int lotId = pair.first;
+    if (currentCars.count(lotId) == 0)
+      return lotId;
+  }
+  return createNewLotId();
+}
+
+int Parking::createNewLotId() {
+  int maxId = currentCars.empty() ? 0 : currentCars.begin()->first + 1;
```
teodoraargintaru: I think it would be nicer to use currentCars.size() here.

group/assignment-6/assignment-6/cars.cpp
```diff
@@ -0,0 +1,99 @@
...
+void Parking::moveCarToTheEmptyLot() {
+  if (rightLotIsEmpty())
+    return;
+  int car = plannedCars[emptyLot], fromLotId = currentLots[car];
+  plannedCars.erase(emptyLot);
+  removeCar(fromLotId);
+  moveCarToTheEmptyLot();
+}
+
+void Parking::removeCar(int fromLotId) {
```
teodoraargintaru: I like the naming stile you used in this method, it is very self-explanatory :)

group/assignment-6/assignment-6/cars.h
```diff
@@ -0,0 +1,64 @@
...
+  bool operator==(const Move &other) const noexcept;
+  friend std::ostream &operator<<(std::ostream &os, const Move &move);
+
+private:
+  int carId, fromLotId, toLotId;
+};
+
+class Parking {
+public:
+  Parking(const std::map<int /*lotId*/, int /*carId*/> &startState,
```
sanjamijovic: Nice comment on what map holds.

group/assignment-6/assignment-6/cars.h
```diff
@@ -0,0 +1,64 @@
...
+  bool findLotWithWrongCar(int *lot);
+  // Returns true if currently empty lot is actually planned to have no cars.
+  bool rightLotIsEmpty();
+  // Returns max of used lots' id  incresed by 1.
+  int createNewLotId();
+  // Tries to deduce the id of empty lot. If it cannot be deduced from the given 
+  // parking states, chooses an unused id.
+  int calculateEmptyLotId();
+  // Fills the currentLots map based on currentCars map.
+  void saveCurrentLots();
```
sanjamijovic: Very well documented!

PR #25, opened 2019-07-19
Sanjamijovic assignment6

sanjamijovic:
Since slots are represented by a letter, I created a mapping from letters to unique ids stored in ParkingLot class. ParkingLotState class represents some state on the parking lot. I stored cars in a vector on the position of the unique id of the parking slot. The way the function rearranges the cars is: if there is some car that should be stored on the current free slot in the end state that that car is moved, otherwise I move the first non-null car from the parking lot. This step is repeated until the end state is reached.


group/assignment-6/ParkingLotState.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+        int spaceToBeMovedFrom = std::distance(parkingSlotStates_.begin(),
+                positionToMoveFrom);
+
+        auto slotFrom = std::find_if(
+                parkingLot_.parkingSpaceIdMapping_.begin(),
+                parkingLot_.parkingSpaceIdMapping_.end(),
+                [spaceToBeMovedFrom](const auto& element)
+                { return element.second == spaceToBeMovedFrom; });
+        auto slotTo = std::find_if(
```
teodoraargintaru: The code kind of repeats itself here. maybe you could have used a function for slotTo and slotFrom?

group/assignment-6/test.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+        for (auto move : moves) {
+            startState[move.toSlot] = startState[move.fromSlot];
+            startState[move.fromSlot] = nullptr;
+        }
+
+        EXPECT_TRUE(startState == endState);
+    }
+
+    TEST(InvalidStates) {
```
teodoraargintaru: Nice test! :)

group/assignment-6/ParkingLotState.cpp
sanjamijovic: Right, I totally agree.

group/assignment-6/ParkingLotState.cpp
```diff
@@ -0,0 +1,76 @@
...
+        // finds the car that should be moved to current free position according
+        // to the end state or first non-null car
+        if (endState[freeSlot_] != nullptr) {
+            positionToMoveFrom = std::find(parkingSlotStates_.begin(),
+                    parkingSlotStates_.end(),
+                    endState[freeSlot_]);
+        } else {
+            positionToMoveFrom = std::find_if(parkingSlotStates_.begin(),
+                    parkingSlotStates_.end(),
+                    [](Car* element) -> bool { return element != nullptr;});
```
annetka-k: I think there we should choose not any not-null car, but a car that is on the wrong slot
Otherwise it will go in infinite loop. For example, for
{ {'A', nullptr}, {'B', car1}, {'C', car2}, {'D', car3} } as startState
{ {'A', nullptr}, {'B', car1}, {'C', car3}, {'D', car2} } as endState
we will always swap car1 from 'B' to 'A' and back.

group/assignment-6/ParkingLot.h
```diff
@@ -0,0 +1,37 @@
...
+// Represents parking lot with slots defined by letter. Each slot in one parking
+// lot has its unique id.
+class ParkingLot {
+    friend class ParkingLotState;
+public:
+    explicit ParkingLot(const std::vector<char>& slotNames);
+
+    inline friend bool operator==(const ParkingLot& first,
+            const ParkingLot& second) {
+        return first.id == second.id;
```
annetka-k: Looks a bit complicated with these ids.
Why not compare just by parkingSpaceIdMapping_?
Otherwise, for 
ParkingLot parkingLot1({ 'A', 'B', 'C' });
ParkingLot parkingLot2({ 'A', 'B', 'C' });
we will get that parkingLot1!= parkingLot2, which is quite illogical.

group/assignment-6/ParkingLotState.cpp
sanjamijovic: Right, I agree.

group/assignment-6/ParkingLot.h
sanjamijovic: Yes, what I intended here is to compare just if they are the same objects. So cars have to have the same object of parking lot passed, but I agree that they can be compared by mapping.

PR #26, opened 2019-07-21
Basic implementation of main problem:

Posiadacz07:
* tests 
* scripts for building and testing
* all documentation


group/assignment-6/rerrangeCars_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+  std::vector<int> initialOrder;
+  std::vector<int> finalOrder;
+
+  for (int i = 0; i < 100; i++) {
+    initialOrder.push_back(i);
+    finalOrder.push_back(100 - 1 - i);
+  }
+  std::vector<std::vector<int>> result =
+      generateParkingSequence(initialOrder, finalOrder);
+  EXPECT_EQ(result.back(), finalOrder);
```
sanjamijovic: Great test!

group/assignment-6/rerrangeCars_test.cpp
```diff
@@ -0,0 +1,68 @@
...
+#include "../../shared/testing/test.h"
+
+#include <iostream>
+
+namespace {
+TEST(EmptyParking) {
+  std::vector<int> initialOrder{0};
+  std::vector<int> finalOrder{0};
+  std::vector<std::vector<int>> result =
+      generateParkingSequence(initialOrder, finalOrder);
```
sanjamijovic: I would recommend returning some structure that represents move so that you can verify moves for ex. if you get to the final order by produced moves.

